Issue #21, opened 2013-10-16
Way to set attachment name?


taylorotwell:
Currently you add an attachment by file path. Is there a way to set a custom name for the attachment that is distinct from the path?

gorgano:
I wasn't able to find any specification in the documentation on how to do this.  In case anyone was looking to do this, it appears you can just pass an array of attachments with the appropriate name fields to get named attachments:

           $attachments[] = [
                'remoteName' => 'nameofthefile.ext', 
                'filePath' => '/path/to/the/file'
            ];

DevNIX:
This feature should be documented

Nyholm:
Thank you @DevNIX. I've created an issue for that. If you want, you are free to create a PR for this.

goaiworld:
> I wasn't able to find any specification in the documentation on how to do this. In case anyone was looking to do this, it appears you can just pass an array of attachments with the appropriate name fields to get named attachments:
> 
> ```
>        $attachments[] = [
>             'remoteName' => 'nameofthefile.ext', 
>             'filePath' => '/path/to/the/file'
>         ];
> ```

It's filename, instead of remoteName in sdk version 3.x

Issue #20, opened 2013-10-15
Put Request on Routes


travelton:
Suspected: Actions seem to be unordered after issuing a PUT request on a Route. Needs more research.

PR #29, opened 2014-01-13
Ability to rename attachments and inline images & Duplicate Post/Put Fields

travelton:
This commit pushes the requirement for Guzzle up one point release to allow the ability to rename files. The SDK has been configured to take advantage of this new feature. 

To send a message, and rename the files, use an array of arrays like the following example: 

```
# First, instantiate the SDK with your API credentials and define your domain. 
$mg = new Mailgun("key-example");
$domain = "example.com";

# Now, compose and send your message.
$mg->sendMessage($domain, array('from'    => 'bob@example.com', 
                                'to'      => 'sally@example.com', 
                                'subject' => 'The PHP SDK is awesome!', 
                                'text'    => 'It is so simple to send a message.'), 
                                array(array('filePath'   => '@/receipts/customer892374892.pdf', 
                                            'remoteName' => 'Receipt.pdf'),
                                      array('filePath'   => '@/taxforms/IRS/1099.pdf', 
                                            'remoteName' => 'IRS-1099.pdf')));
```

Alternatively, you can use MessageBuilder to handle this:

```
# Add an attachment
addAttachment('@/receipts/customer892374892.pdf', 'Receipt.pdf');

# Add an inline file
addInlineImage('@/taxforms/IRS/1099.pdf', 'IRS-1099.pdf');
```

This backward compatibility code should be removed in the next major release. 

Refs #21 

The second commit here includes a fix for post/put field aggregation. Same keys should be duplicated, much like Python/Ruby style HTTP payloads. This fix shouldn't affect existing code.

Refs #20

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 Enhancement:
   - adjust file attachment/inline name (#21 @travelton)
 
+Bugfixes:
+  - fixed issue with unordered route actions (#23 @travelton)
+
 ## 1.5 (2013-12-13)
 
 Enhancement:
--- a/src/Mailgun/Connection/RestClient.php
+++ b/src/Mailgun/Connection/RestClient.php
@@ -6,6 +6,7 @@ use Guzzle\Http\Client as Guzzle;
 use Mailgun\MailgunClient;
 
 use Mailgun\Connection\Exceptions\GenericHTTPError;
+use Guzzle\Http\QueryAggregator\DuplicateAggregator;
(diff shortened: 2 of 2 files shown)


murtraja:
i discovered a bug here
if you have remote name like "my file on 30/8/2016", then only 2016 is considered

Nyholm:
Thank you @murtraja for reporting this. This is a very old PR, would you mind creating a new issue for this? Or even better, add a pull request with a failing test.

PR #74, opened 2015-03-15
Allow endpoints with starting /

jenssegers:
The first thing I did with the API was `$mailgun->get('/domains')`, which returned a `MissingEndpoint` exception. I think both uri's should be possible, both `domains` and `/domains`.


Nyholm:
I don't understand why you would want this. Why can't you just call the `$mailgun->get()` with valid parameters?
I think it is unnecessary to run an extra `ltrim` on every API request when we don't need to. 

:-1:

PR #86, opened 2015-07-23
Bumped to guzzlehttp/guzzle:~6.0

auro1:
DO NOT MERGE YET

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
     "name": "mailgun/mailgun-php",
     "description": "The Mailgun SDK provides methods for all API functions.",
     "require": {
-        "guzzlehttp/guzzle": "~5.0"
+        "guzzlehttp/guzzle": "~6.0"
     },
     "require-dev": {
         "php": ">=5.4.0",
--- a/src/Mailgun/Connection/RestClient.php
+++ b/src/Mailgun/Connection/RestClient.php
@@ -3,16 +3,15 @@
 namespace Mailgun\Connection;
 
 use GuzzleHttp\Client as Guzzle;
-use GuzzleHttp\Message\ResponseInterface;
-use GuzzleHttp\Post\PostBodyInterface;
-use GuzzleHttp\Post\PostFile;
-use GuzzleHttp\Query;
+use GuzzleHttp\Psr7\Request;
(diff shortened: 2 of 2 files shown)


auro1:
Travis failed - because one of the instances couldn't find version 6.
Don't merge just yet; I have a feeling the PHPUnit tests are not completely covering it all.

I'm manually testing it as we speak.

auro1:
Refactoring required. DO NOT MERGE THIS PR YET
I'm working on upgrading the rest client to Guzzle/6.0 and adding new phpunit tests to ensure they cover the changes.

https://github.com/guzzle/guzzle/blob/master/UPGRADING.md

auro1:
Still onto this, I'm currently very busy in a paid project, but I will get back to finish this up with tests. The build I have is functional; it just dosn't have solid tests.

auro1:
Is it a problem that it's not compatible with PHP 5.4 ?

Nyholm:
May I suggest we decuple from the HTTP implementation by using PHP-HTTP. Read about it here: http://php-http.readthedocs.org/en/latest/

I'm happy to help if you want.

RubenKruiswijk:
@Nyholm Imho, when adopting http://www.php-fig.org/psr/psr-7/ Message Interfaces we should be able to implement any library that supports it, including the new Guzzle 6. Migrating to Guzzle 6 first could make that happen more easily I think.

Nyholm:
> Ignore my comment above - http://php-http.readthedocs.org/en/latest/ includes a lot more, including PSR-7

Yeah, I'm glad you checked it out. I've not done such implementation myself. But I believe that is needed because this library should not care if the developer are using Guzzle5, Guzzle6 or any other transport. `php-http/adapter` abstract away that problem and we will not longer bother with implementing a new Guzzle version.

bshaffer:
I would suggest making PHP 5.5 the minimum version (since php 5.4 is EOL'd, and still supported through `v1.8` of this library) and merging this request.

bshaffer:
... or merging #94. Either way, would be great to see Guzzle 6 supported.

sagikazarmark:
> ... or merging #94.

:+1:

PR #93, opened 2015-10-01
Add function to configure Guzzle's default options, so the user can set the Guzzle timeout.

dg192:
I have added a function to RestClient.php which would allow the user to configure the default options that the Guzzle client uses.  I then added a function to Mailgun.php which would let the user pass in an array of Guzzle options.  See the README file for an example.

The purpose for this, is that by default, Guzzle requests wait indefinitely.  This was causing my application to halt, because for some reason I wasn't getting a response when making requests.

As a side note, I believe it would be a good idea to add default values for "timeout" and "connect_timeout" in the RestClient.php constructor.  However, I think that might be considered a "breaking change" so I didn't add it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,22 @@ $mg->get("$domain/log", array('limit' => 25,
                               'skip'  => 0));
 ```
 
+Guzzle Configuration
+--------------------
+
+Mailgun uses Guzzle to handle HTTP requests.  Here's how you would configure the default options
+for Guzzle...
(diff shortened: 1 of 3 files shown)


bfrohs:
:+1:

Nyholm:
This issue will be fixed by #94. You could then configure the client however you want.

Nyholm:
Closed by #94

PR #88, opened 2015-09-09
upgrade guzzle version

farmani:
you need to update guzzle version to ^6.0

PR #94, opened 2015-10-03
Remove dependency on Guzzle 

Nyholm:
In order to not be dependent on any version of Guzzle I've implemented [PHP-HTTP](http://php-http.readthedocs.org/en/latest/). This PR will replace #88 and #86. Please try and review carefully. 

***Don't merge this PR until ´php-http` is stable.*** It is stable now (2016-02-28)

I use Guzzle's psr7 package to properly encode the multipart. 

Ping @auro1, @RubenKruiswijk and @farmani any 

---
Question: What is the story behind `CURLOPT_FORBID_REUSE`? I can't find anything about it. 
Answer: it was used because we wanted to help the load balancers to distribute the load. The load balancers used today are more clever and this setting is not needed.


composer.json
```diff
@@ -2,11 +2,13 @@
     "name": "mailgun/mailgun-php",
     "description": "The Mailgun SDK provides methods for all API functions.",
     "require": {
-        "guzzlehttp/guzzle": "~5.0"
+        "guzzlehttp/psr7": "~1.2",
```
auro1: Why not php-fig/http-message ?

Nyholm: You are correct. We should not be dependent on Guzzle's implementation here. We should use a message factory.. I'll update in a minute.

Nyholm: Sorry, We need the implementation for because there are too difficult to handle file upload without it. See here: https://github.com/mailgun/mailgun-php/pull/94/files#diff-72d96caf4bb8f89a5d30eafdff75bbe2R63

composer.json
```diff
@@ -2,11 +2,13 @@
...
+        "happyr/http-auto-discovery": "0.1.*"
```
RubenKruiswijk: What about using only ``php-http/discovery``? IMHO we should be careful adding dependencies. As noted in the readme of happyr/http-auto-discovery, this can be replaced with only a few lines of code. (I noticed you are the author of the library, so you probably wrote that yourself :-))

https://github.com/Happyr/HttpAutoDiscovery

Nyholm:
First of all. Don't merge this until php-http has reached a stable version. (I'll add that in the PR's description). 

@RubenKruiswijk  `php-http/discovery` is used to find a transportation library and to find a factory to create requests. 
Our requests we use here are quite complex (with the file upload). It is too difficult to make sure we got all the parameters right etc etc, so We should use a library that do that for us. This is not a part of PSR-7 btw. 
Guzzle's implementation helps us with that (there are probably others as well). 

@RubenKruiswijk Do you think I should remove `happyr/http-auto-discovery`? It was a good fit in the beginning but the complexity of file upload removed the beauty and simplicity of using `happyr/http-auto-discovery`.

Nyholm:
I removed dependency on `happyr/http-auto-discovery` and using `php-http` directly.

src/Mailgun/Connection/RestClient.php
```diff
@@ -25,30 +25,59 @@ class RestClient
...
-    public function __construct($apiKey, $apiEndpoint, $apiVersion, $ssl)
+    public function __construct($apiKey, $apiHost, $apiVersion, $ssl, HttpAdapter $adapter = null)
     {
         $this->apiKey = $apiKey;
-        $this->mgClient = new Guzzle([
-            'base_url'=>$this->generateEndpoint($apiEndpoint, $apiVersion, $ssl),
-            'defaults'=>[
-                'auth' => array(Api::API_USER, $this->apiKey),
-                'exceptions' => false,
-                'config' => ['curl' => [ CURLOPT_FORBID_REUSE => true ]],
```
RubenKruiswijk: We need a way to set these defaults (CURLOPT_FORBID_RESUSE and User-Agent header)

src/Mailgun/Connection/RestClient.php
```diff
@@ -25,30 +25,59 @@ class RestClient
...
+     */
+    protected $apiEndpoint;
 
     /**
      * @param string $apiKey
-     * @param string $apiEndpoint
+     * @param string $apiHost
      * @param string $apiVersion
      * @param bool   $ssl
      */
```
RubenKruiswijk:
```php
@param HttpAdapter $adapter
```

src/Mailgun/Connection/RestClient.php
```diff
@@ -25,30 +25,59 @@ class RestClient
...
+        if ($this->httpAdapter === null) {
+            $this->httpAdapter = HttpAdapterDiscovery::find();
+        }
+
+        $headers['User-Agent'] = Api::SDK_USER_AGENT.'/'.Api::SDK_VERSION;
+        $headers['Authorization'] = 'Basic '.base64_encode(sprintf('%s:%s', Api::API_USER, $this->apiKey));
+
+        if (!empty($files)) {
+            $body = new MultipartStream($files);
+            $headers['Content-Type'] = 'multipart/form-data; boundary='.$body->getBoundary();
```
RubenKruiswijk: body can be an empty array?

src/Mailgun/Connection/RestClient.php
Nyholm: I understand that we need to set `CURLOPT_FORBID_RESUSE` for backwards compatibility, but why is that needed? I cant find any documentation about why we use it. This is from the [official documentation](http://curl.haxx.se/libcurl/c/CURLOPT_FORBID_REUSE.html)

> Pass a long. Set close to 1 to make libcurl explicitly close the connection when done with the transfer. Normally, libcurl keeps all connections alive when done with one transfer in case a succeeding one follows that can re-use them. This option should be used with caution and only if you understand what it does as it can seriously impact performance.

> Set to 0 to have libcurl keep the connection open for possible later re-use (default behavior).

src/Mailgun/Connection/RestClient.php
Nyholm: Thank you!

src/Mailgun/Connection/RestClient.php
Nyholm: I guess it can, but the default is null. I'll better change

src/Mailgun/Connection/RestClient.php
Nyholm: Here is the commit that added the CURLOPT_FORBID_RESUSE.
https://github.com/mailgun/mailgun-php/commit/b4b9f0ec4018db76670e0beceff9e1021880d58c

Ping @travelton, I know it was over 2 years ago, but do you remember why you added that?

travelton: @Nyholm If I recall, it was an internal decision to help force connections between API nodes, thus, improving the overall performance from a customer's perspective.   

For example, if you are processing 1,000 API calls, reusing connections, and you land on a node that isn't performing very well, you'd issue 1,000 API calls on that non-performant node. With this option enabled, you effectively force a new connection to a different API node every time. So now, maybe one in every x API calls hits a non-performant API node.  

I'm betting this "trick" is obsolete, as the load balancing architecture has rapidly improved/changed over the years. Unfortunately, I don't have the insight to confirm, as I'm not working with the Mailgun team these days. As an outsider looking in, I would remove it from default configurations, and allow it as an "Advanced Option" for tinker-ers that love performance tuning. :)

Nyholm: Thank you for your input. Do you know who are on the Mailgun team that can confirm this?

travelton: @Nyholm Reached out to the Mailgun team, they confirmed this setting will have no performance benefit due to improvements in load balancing.

Nyholm: Thank you @travelton. You are awesome!

src/Mailgun/Connection/RestClient.php
Nyholm: Oh, you mean at line 74? No, it gets a new value on line 73.

Nyholm:
PHP-HTTP is probably stable now. They will tag a stable version before the end of the year. I've updated this PR to reflect the latest changes in PHP-HTTP.

virgofx:
Looking forward to this +1

bshaffer:
+1 on this! What else needs done to get this merged and tagged?

captn3m0:
:+1: from me as well.

Nyholm:
Thank you @bshaffer 
We are waiting on Httplug to get stable. It is currently in beta and discussing release candidates. As soon as it get stable I'll update this PR and let you know.

captn3m0:
httplug had its first stable release a while back: https://github.com/php-http/httplug/releases/tag/v1.0.0

Nyholm:
Yeah, they did. :smile:  The discovery repo is not stable yet though. I am following the discussions in Httplug and in Puli. 

I can do some updates here to make sure we are closer to a merge.

tomschlick:
Any news on this? 

The fiasco over at Mandrill has caused me to look to Mailgun for our email platform and the dependency on guzzle v5 is conflicting with my existing codebase that has v6.

virgofx:
@tomschlick You can temporarily use the following fork until everything is merged.

```
"repositories": [{
        "type": "vcs",
        "url": "https://github.com/MunkSoftware/mailgun-php"
    }],
    "require": {
        "php": ">=5.6.0",
        "guzzlehttp/guzzle": "~6.0",
        "mailgun/mailgun-php": "dev-master#04f1baa2fdbb4d3fba035e28ce66ec676e382523",
        ...
   "
}]
```

tomschlick:
Don't know how I didn't think of that. I've done it many times before...
Thanks!

--
Tom Schlick




On Thu, Feb 25, 2016 at 4:55 PM -0800, "Mark Johnson" <notifications@github.com> wrote:












@tomschlick You can temporarily use the following fork until everything is merged.

"repositories": [{
        "type": "vcs",
        "url": "https://github.com/MunkSoftware/mailgun-php"
    }],
    "require": {
        "php": ">=5.6.0",
        "guzzlehttp/guzzle": "~6.0",
        "mailgun/mailgun-php": "dev-master#04f1baa2fdbb4d3fba035e28ce66ec676e382523",
        ...
   "
}]

Nyholm:
I has been almost 5 months since I've opened this PR and it is now ready for a final review. Please be extra observant to the readme and the new installation instructions. Do they make sense? Do you understand that you need to do one of the following: 
1) create a instance of a HttpClient and inject to Mailgun 
2) install Puli. 

After reviewing and fixed the issues you have found I will squash my commits and make sure it is ready for merge.

Nyholm:
Ping @virgofx @captn3m0 @tomschlick @bshaffer @jessespears Do you have any input?

tomschlick:
```php
$mailgun = new \Mailgun\Mailgun('api_key', null, null, true, $client);
```

This is a little crazy with 5 params, 3 of which most people wouldn't touch.

Since this is going to be a breaking change, can we get rid of some of the options in construct and throw some of them to their own methods?

For instance if we got rid of the SSL = true and the api version it would look much cleaner as 

```php
$mailgun = new \Mailgun\Mailgun('api_key', 'mydomain.com', $client);

// If you needed those options you could do...

$mailgun->setApiVersion('v2'); // default current version
$mailgun->setSSLEnabled(false); // default true
```

virgofx:
Agree with @tomschlick 's comment. 

@Nyholm I'll review the changes this weekend and provide some feedback.

Nyholm:
Thank you for the review. I agree, the constructor was very bloated with parameters. I removed API version and SSL as you suggested.

tomschlick:
:+1: looks good!

src/Mailgun/Mailgun.php
```diff
@@ -29,13 +30,16 @@ class Mailgun{
...
-    public function __construct($apiKey = null, $apiEndpoint = "api.mailgun.net", $apiVersion = "v3", $ssl = true){
+    public function __construct(
+        $apiKey = null,
+        HttpClient $httpClient = null,
+        $apiEndpoint = 'api.mailgun.net'
+    ) {
         $this->apiKey = $apiKey;
-        $this->restClient = new RestClient($apiKey, $apiEndpoint, $apiVersion, $ssl);
+        $this->restClient = new RestClient($apiKey, $apiEndpoint, $httpClient);
     }
```
virgofx: 1. Can we make $apiKey required? I don't think you can use Mailgun without it?

README.md
```diff
@@ -141,7 +161,9 @@ Go to http://bin.mailgun.net. The Postbin will generate a special URL. Save that
 
 ```php
 # First, instantiate the SDK with your API credentials and define your domain. 
-$mg = new Mailgun('key-example', 'bin.mailgun.net', 'aecf68de', $ssl = False);
+$mg = new Mailgun('key-example', null, 'bin.mailgun.net');
+$mg->setApiVersion('aecf68de');
+$mg->setSslEnabled('false');
```
virgofx: Should probably have an example that is closer to how it's used with mailgun. What is the 'bin.mailgun.net'? 

Most users would do something similar:

```
$mg = new Mailgun('your-key');

....
```
And that should be it. That way SSL is enabled by default and less confusion.

virgofx:
Looks good to me. Only comment was updating the Readme to not have the null options in construct as the first example ... if you wanted to do that .. perhaps make two examples. The first one should always be the simple default for how most people can use the API.  The second one could be with a different rest client and different endpoint.  And ... wasn't sure was it possible to have a null API key? If it's not, make this parameter not optional.

README.md
Nyholm: This is the [existing example](https://github.com/mailgun/mailgun-php#debugging) of how to debug by posting the output to pastebin. This should later be updated, example by using the [Logger plugin](https://github.com/php-http/plugins/blob/master/src/LoggerPlugin.php).

src/Mailgun/Mailgun.php
Nyholm: I agree, API key should be required. I did not want to make too many BC breaking changes. That's why I left is as it was. 

Thank you for your review.

mollierobbert:
When will this be merged?

tristanbes:
Interested in this getting merged ASAP too ;-)

Nyholm:
I have contacted @jessespears and he said he will do his best to have a look at it before the weekend.

jessespears:
Thanks for this contribution, @Nyholm!  This is great!

sagikazarmark:
Nice Job @Nyholm

PR #100, opened 2015-12-29
Update RestClient.php

firemankurt:
The comparison does not need to be "if ($httpResponseCode === 200)".  This fails because $httpResponseCode may come back as a string.
This caused error in my Laravel installation. 

"if ($httpResponseCode == 200)" is all that is required.


Nyholm:
One should never user week comparison. But the issue you are referring to will be fixed by #94

Nyholm:
This can issue can be closed. It is no longer a problem in 2.0

PR #107, opened 2016-02-26
Updated minimum PHP version

RubenKruiswijk:
Because the library uses the short array syntax ``[]``, vs ``array()``.


RubenKruiswijk:
My bad, that was already possible in 5.4.

PR #91, opened 2015-09-25
Updated README version number

lahdekorpi:

PR #108, opened 2016-03-02
Updated composer to pull latest version 1.8

mistylackie:
The documentation was instructing to pull 1.7.2 but the current version is 1.8. Some things such as bounces have errors on older versions so it would be better for people to download 1.8 to avoid issues with the latest features.

It would be great if the Library Download link (line 30) was updated too but I do not know the link to 1.8

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ composer and the Mailgun SDK.
 curl -sS https://getcomposer.org/installer | php
 
 # Add Mailgun as a dependency
-php composer.phar require mailgun/mailgun-php:~1.7.2
+php composer.phar require mailgun/mailgun-php:~1.8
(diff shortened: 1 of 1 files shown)


Nyholm:
Good suggestion. But maybe remove the version completely? If so, Composer will fetch the latest version.

mistylackie:
Yes removing the version would be even better.

Nyholm:
For the record: This is a duplicate of #91.

obukhov-sergey:
@mistylackie @Nyholm thanks for the PR and your interest in the project. The README has already been updated to point to 2.0 release. I also like the idea about removing the version completely. Will try to submit a PR soon unless there is one already. Closing this one.

PR #114, opened 2016-03-31
Allow guzzle v6 as well

bfrohs:


sagikazarmark:
I this became invalid

sagikazarmark:
Not to mention Guzzle 5 and 6 work totally different.

PR #129, opened 2016-05-08
Update v2 to v3

marco258:
Update the API version from v2 to v3 on PHP-SDK 2.0


Nyholm:
Are you sure that there is no other changes that are needed than this?

obukhov-sergey:
Thanks for your interest and the PR but some endpoints still support only v2 and not v3 e.g. ``/v2/<domain name>/bounces``. To change the version you can always do

```php
$mg = new Mailgun('key-example', null, 'bin.mailgun.net');
$mg->setApiVersion('v3');
```

Issue #122, opened 2016-04-23
Unable to put() with any data


ianchadwick:
Attempting the below I get "Invalid resource type: array".

```php
$mailgun->put('domains/' . $domain . '/connection', [
    'require_tls' => true,
    'skip_verification' => true,
]);
```

The only way I've been able to update the item is to use 'domains/' . $domain . '/connection?require_tls=true' which is a bit hacky.

Also if you don't pass anything as the 2nd param it complains due to missing argument.

nisbeti:
Since the upgrade to v2 I have the same problem, which I have managed to (hack) fix, until a proper fix is released. I am using 
```php
$client = new \Http\Adapter\Guzzle6\Client();
```
as I didn't want to go the Puli discovery route.

./vendor/guzzlehttp/psr7/src/functions.php has a function called ```stream_for``` at line 77, which checks the type of the $resource passed to it (in our case this is the $body of our PUT request), and rejects it if it is an array.

Either we must send the $body as a urlencoded string, ```'name=Bob&subscribed=false'``` along with setting a content-type header of ```application/x-www-form-urlencoded``` (which means hacking the mailgun RestClient for now), or we can stick with the array of params, but then we still have to hack to RestClient in order to convert the array into a urlencoded string and set the header.

Here is the hack for file ```./vendor/mailgun/mailgun-php/src/Mailgun/Connection/RestClient.php```

*This is Strictly Temporary, as it will be overwritten next time mailgun-php is updated*

At line 86 (just before ```$request = new Request($method, $this->getApiUrl($uri), $headers, $body);``` insert:
```php
        if (is_array($body)) {
            $headers['Content-Type'] = 'application/x-www-form-urlencoded';
            $urlencodedBody = '';
            foreach ($body as $key => $value) {
                $urlencodedBody .= urlencode($key) . '=' . urlencode($value) . '&';
            }
            $body = trim($urlencodedBody,'&');
        }
```

A more permanent fix may include being able to send the content-type header from the PUT method, but I am unsure what impact that will have when not using the Guzzle6\Client and using Puli discover instead.

iRbouh:
First, make sure you are using the last version of `php-http/guzzle6-adapter`, then try the following:

```
$client = \Http\Adapter\Guzzle6\Client::createWithConfig([
            'headers' => [
                'Content-Type' => 'application/x-www-form-urlencoded'
            ]
        ]);
[...]

PR #131, opened 2016-05-11
add header and convert arrays for put method

nisbeti:
Re: Issue https://github.com/mailgun/mailgun-php/issues/122

Since the upgrade to v2 and using ```$client = new \Http\Adapter\Guzzle6\Client();```
```./vendor/guzzlehttp/psr7/src/functions.php``` has a function called ```stream_for``` at line 77, which checks the type of the $resource passed to it (in our case this is the $body of our PUT request), and rejects it if it is an array. We can convert the array to send the $body as a urlencoded string, 'name=Bob&subscribed=false' along with setting a content-type header of ```application/x-www-form-urlencoded```

### DIFF
--- a/src/Mailgun/Connection/RestClient.php
+++ b/src/Mailgun/Connection/RestClient.php
@@ -182,7 +182,16 @@ class RestClient
      */
     public function put($endpointUrl, $putData)
     {
-        return $this->send('PUT', $endpointUrl, $putData);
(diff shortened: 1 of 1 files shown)


Nyholm:
This solution is bit hacky and does not fix the same issue for the POST request. See an alternative solution #138

PR #133, opened 2016-05-22
Add new package in Framework integration section

narendravaghela:
I have created a plugin for CakePHP to send emails using Mailgun SDK and added the reference to package in readme file.


narendravaghela:
Hello there,
Is there any possibility to merge this request?

Nyholm:
Your library has in total 10 installs. I would prefer a more commonly used integration. There are cakePHP integrations with far more installation than yours. 

If it was up to me I would be happy to include your library when it has gotten some attraction.

narendravaghela:
Thank you for reply.
If you can add this to list, It would be helpful to get some attraction to this lib. Anyways, not a problem, let's wait until my library gets some more users.

Nyholm:
Yes, lets wait a while.

PR #139, opened 2016-07-19
Do not depend on Guzzle/PSR7

Nyholm:
This will make sure we are totally decoupled from Guzzle. It will also make installation a bit easier since php-http/discovery:1.0 does no longer depend on Puli.


Nyholm:
Thank you for merging!

PR #151, opened 2016-08-05
Fixed inline/attachment issue. Added new tests.

travelton:
Hello Friends!

This pull request fixes a multi inline/attachment issue. Some HTTP clients cannot handle url encoded form data with the same key and different value. With the current implementation, if you pass in multiple inline/attachment values, only the last value in the array is actually sent to the Mailgun API.

We need to give the key an "index marker" (for lack of better term?). The Mailgun API knows how to handle this...

Example (Bad):
```
inline=blah&inline=blah
```

Example (Good):
```
inline[0]=blah&inline[1]=blah
```

Also, tests. I'm including my lame attempt at revamping the tests a bit. Eventually, every test should probably follow something like this, so as to test between request and "$client->send()". Happy to entertain any community suggestions here. We really need to beef up the tests on this SDK.

### DIFF
--- a/src/Mailgun/Connection/RestClient.php
+++ b/src/Mailgun/Connection/RestClient.php
@@ -119,8 +119,10 @@ class RestClient
         foreach ($fields as $fieldName) {
             if (isset($files[$fieldName])) {
                 if (is_array($files[$fieldName])) {
+                    $fileIndex = 0;
                     foreach ($files[$fieldName] as $file) {
-                        $postFiles[] = $this->prepareFile($fieldName, $file);
+                        $postFiles[] = $this->prepareFile($fieldName, $file, $fileIndex);
+                        $fileIndex++;
                     }
                 } else {
                     $postFiles[] = $this->prepareFile($fieldName, $files[$fieldName]);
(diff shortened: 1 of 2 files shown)


src/Mailgun/Connection/RestClient.php
```diff
@@ -251,10 +253,11 @@ protected function getResponseExceptionMessage(ResponseInterface $responseObj)
      *
      * @param string       $fieldName
      * @param string|array $filePath
+     * @param integer      $fileIndex
      *
      * @return array
      */
-    protected function prepareFile($fieldName, $filePath)
+    protected function prepareFile($fieldName, $filePath, $fileIndex=0)
```
Nyholm: Code style `$fileIndex = 0`

Nyholm:
Thank you @travelton. This is a great PR. 

Im not happy with how you write the tests. I think that is a hack. But your point is valid. I'll send a PR shortly that achieve the same goals with the tests. Give me a moment and I'll have something to show

Nyholm:
See #153. 
I was planning to do something like the code example in the PR but doing that would be way to much boilerplate code for each test.

travelton:
@Nyholm Agreed, totally hacky. Removing my tests. Once you have your tests merged, I'll rebase this PR and write some tests for it.

Nyholm:
Awesome. Im glad you liked it. I've merged the #153 now

Nyholm:
Merged by #156

PR #158, opened 2016-08-12
Add attachment from memory

wimverstuyf:
Changed RestClient so an attachment can be added from memory.
Instead of the filepath an array with 'filename' and 'fileContent' can be used.
Needed this functionality because the attachment is generated on-the-fly in memory.


Nyholm:
Thank you. This is great! 
I would be happy to merge this PR but it is missing tests. Would you like to add one? 

Use this functional tests as a template: https://github.com/mailgun/mailgun-php/blob/master/tests/Functional/SendMessageTest.php

wimverstuyf:
Great!
I've added the unit test.

tests/Functional/FileFromMemoryTest.php
```diff
@@ -0,0 +1,38 @@
...
+            $this->assertContains(['name'=>'to',      'contents'=>'alice@example.com'], $files);
+            $this->assertContains(['name'=>'subject', 'contents'=>'Foo'], $files);
+            $this->assertContains(['name'=>'text',    'contents'=>'Bar'], $files);
+            $this->assertContains(['name'=>'attachment[0]', 'contents'=>'First file content', 'filename' => 'file1.txt'], $files);
+            $this->assertContains(['name'=>'attachment[1]', 'contents'=>'Second file content', 'filename' => 'file2.txt'], $files);
+        };
+
+        $attachments = [
+            ['filename' => 'file1.txt', 'fileContent' => 'First file content'],
+            ['filename' => 'file2.txt', 'fileContent' => 'Second file content']
```
Nyholm: Let's add the second file by a path to make sure both works

tests/Functional/FileFromMemoryTest.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+class FileFromMemoryTest extends \PHPUnit_Framework_TestCase
+{
+    public function testAddFileFromMemory()
+    {
+        $fileValidator = function($files) {
+            $this->assertContains(['name'=>'from',    'contents'=>'bob@example.com'], $files);
+            $this->assertContains(['name'=>'to',      'contents'=>'alice@example.com'], $files);
+            $this->assertContains(['name'=>'subject', 'contents'=>'Foo'], $files);
+            $this->assertContains(['name'=>'text',    'contents'=>'Bar'], $files);
```
Nyholm: You could remove these extra assert that are not related to attachments.

Nyholm:
What would happens if fileContent is not a string? Do we need a check for that?

tests/Functional/FileFromMemoryTest.php
```diff
@@ -0,0 +1,51 @@
...
+ * @author Wim Verstuyf <wim.verstuyf@codelicious.be>
+ */
+class FileFromMemoryTest extends \PHPUnit_Framework_TestCase
+{
+    public function testAddFileFromMemory()
+    {
+        $fileValidator = function($files) {
+            $fileChecks = [
+                ['name'=>'attachment[0]', 'contents'=>'File content', 'filename' => 'file1.txt'],
+                ['name'=>'attachment[1]', 'filename' => 'mailgun_icon1.png']
```
Nyholm: You have to test the contents of attachment[1], right? Create a new fixture text file and make sure the contents match the expected values.

wimverstuyf:
Changed the unit test as requested:
 - removed unnecessary checks
 - second attachment added from filesystem

fileContent is converted to a Guzzle stream (which calls __toString()) so when it is not a string it will be converted to a string or an error is thrown when __toString() is not available. So this shouldn't be a problem, right?

Nyholm:
> so when it is not a string it will be converted to a string or an error is thrown when __toString() is not available. So this shouldn't be a problem, right?

Correct, or partly correct. The `__toString()` will always be there as it is a part of PSR-7. So you will never get an exception. It will always work. Good!

Did you see this? https://github.com/mailgun/mailgun-php/pull/158#discussion_r77588252

wimverstuyf:
Added a text file for the unit test and reading the content now so this can be compared in the test.

tests/Functional/FileFromMemoryTest.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        $fileValidator = function($files) {
+            $attachments = array_filter($files, function($f) {
+                return strpos($f['name'], 'attachment') !== false;
+            });
+
+            $attachments = array_map(function($f) {
+                return [
+                    'name' => $f['name'],
+                    'contents' => is_resource($f['contents'])?fread($f['contents'], 50):$f['contents'],
```
Nyholm: Will this ever be true? Will $f['contnets'] ever be a resource?

tests/Functional/FileFromMemoryTest.php
```diff
@@ -0,0 +1,46 @@
...
+ * Add attachment with file from memory
+ *
+ * @author Wim Verstuyf <wim.verstuyf@codelicious.be>
+ */
+class FileFromMemoryTest extends \PHPUnit_Framework_TestCase
+{
+    public function testAddFileFromMemory()
+    {
+        $fileValidator = function($files) {
+            $attachments = array_filter($files, function($f) {
```
Nyholm: Smart! 
Write a comment here that explains why you are doing this. 

`// Get only the attachments`

tests/Functional/FileFromMemoryTest.php
wimverstuyf: $f['contents'] is a resource when adding a filepath (fopen is used).

tests/Functional/FileFromMemoryTest.php
wimverstuyf: Okay, added some comments

tests/Functional/FileFromMemoryTest.php
Nyholm: That correct... Hm that is weird.. I do not think we should allow the 'contents' to be both a resource and string. 

We should expect only resources here to at least have a decent API.

src/Mailgun/Connection/RestClient.php
```diff
@@ -261,23 +261,33 @@ protected function getResponseExceptionMessage(ResponseInterface $responseObj)
...
-        if (is_array($filePath)) {
-            $filename = $filePath['remoteName'];
-            $filePath = $filePath['filePath'];
-        }
+        $fileContent = null;
+
+        if (is_array($filePath) && isset($filePath['fileContent'])) {
+            // File from memory
+            $filename = $filePath['filename'];
+            $fileContent = $filePath['fileContent'];
```
Nyholm: Do: 

```php
$resource = fopen('php://temp', 'r+');
fwrite($resource, $filePath['fileContent']);
```

src/Mailgun/Connection/RestClient.php
```diff
@@ -261,23 +261,33 @@ protected function getResponseExceptionMessage(ResponseInterface $responseObj)
...
+
+            // Remove leading @ symbol
+            if (strpos($filePath, '@') === 0) {
+                $filePath = substr($filePath, 1);
+            }
 
-        // Remove leading @ symbol
-        if (strpos($filePath, '@') === 0) {
-            $filePath = substr($filePath, 1);
+            $fileContent = fopen($filePath, 'r');
```
Nyholm: Rename `$fileContent` to `$resource`

src/Mailgun/Connection/RestClient.php
```diff
@@ -261,23 +261,33 @@ protected function getResponseExceptionMessage(ResponseInterface $responseObj)
...
         }
 
         // Add index for multiple file support
         $fieldName .= '[' . $fileIndex . ']';
 
         return [
             'name'     => $fieldName,
-            'contents' => fopen($filePath, 'r'),
+            'contents' => $fileContent,
```
Nyholm: Rename $fileContent to $resource

tests/Functional/FileFromMemoryTest.php
```diff
@@ -0,0 +1,47 @@
...
+{
+    public function testAddFileFromMemory()
+    {
+        $fileValidator = function($files) {
+            // Get only the attachments so the properties can be converted to a format we like
+            $attachments = array_filter($files, function($f) {
+                return strpos($f['name'], 'attachment') !== false;
+            });
+
+            $attachments = array_map(function($f) {
```
Nyholm: `// Convert resources to strings`

Nyholm:
Good tests! They showed the different types of the 'contents'. I think we should only do resources to be consistent. I've made a few comments. 

I think this is all good after that.

Nyholm:
Awesome, thank you!

lahdekorpi:
I really need this functionality and am currently using dev-master as this isn't available on any stable tags.
Is a 2.x coming before the 3.0 release or what should we do now if the 3.0 release is "at least a year away"?

Nyholm:
2.x will come shortly. We just have to make sure we are fine with the new classes. 
I could put some effort in this. Hopefully this weekend.

PR #160, opened 2016-08-22
Add license

clhammer:
Add license + license info to composer.json

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,191 @@
+Apache License
+Version 2.0, January 2004
+http://www.apache.org/licenses/
+
+TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+1. Definitions.
+
+"License" shall mean the terms and conditions for use, reproduction, and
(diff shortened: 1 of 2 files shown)


Nyholm:
Thank you for this PR. We do need a license file. Though, you cannot change the license of an open source project like this. It violates the current license. You need permissions from every contributor. 

Also, why would you change the license to Apache2? I do not see a case where many patents will come out of this project.

clhammer:
Nyholm,

It's my understanding that this wouldn't be a violation as there isn't any actual license declared, just a mention of it in the composer.json meta-data.

Also, I changed it to Apache2 as a move to have license standardization across our SDK's.

Nyholm:
Just because it was a poor license declaration (missing license file and license in the file headers) doesn't mean that there were no license at all. 

Having the same license on all your SDKs makes sense. But why Apache2?

Nyholm:
It has been a while since this PR was updated. 
I suggest closing it in favor for #163. @clhammer what do you think?

Nyholm:
Bump @clhammer

Nyholm:
Closed by #163

PR #172, opened 2016-09-25
Remove php-http/message from installation guide, it's already in deps

limonte:
No need for users to install `php-http/message` manually since it's [already in package deps list](https://github.com/mailgun/mailgun-php/blob/master/composer.json#L8)


Nyholm:
True. However, we **should not** have php-http/message in the composer.json but removing it would be a BC break. 

This PR should be reverted before next major version of the package. .

Nyholm:
Thank you @limonte

PR #118, opened 2016-04-11
Make sure $variables is an array in addRecipient()

lahdekorpi:

PR #176, opened 2016-09-27
Make sure $variables is an array in addRecipient()

DavidGarciaCat:
This PR is replacing the old one (https://github.com/mailgun/mailgun-php/pull/118) providing the expected PHPUnit test to validate the update, and make sure we get a valid response when we don't provide an array with the first name and last name.

This PR is a request of @Nyholm asking for a Pull Request that has the expected tests too.

### DIFF
--- a/tests/Messages/BatchMessageTest.php
+++ b/tests/Messages/BatchMessageTest.php
@@ -41,6 +41,20 @@ class BatchMessageTest extends \Mailgun\Tests\MailgunTestCase
         $this->assertEquals(1, $array['recipients']['to']);
     }
 
+    public function testAddRecipientWithoutFirstAndLastName()
+    {
+        $message = $this->client->BatchMessage($this->sampleDomain);
+        $message->addToRecipient('test@samples.mailgun.org', 'this-is-not-an-array');
+        $messageObj = $message->getMessage();
+        $this->assertEquals(['to' => ["test@samples.mailgun.org"]], $messageObj);
+
+        $reflectionClass = new \ReflectionClass(get_class($message));
+        $property = $reflectionClass->getProperty('counters');
+        $property->setAccessible(true);
+        $array = $property->getValue($message);
+        $this->assertEquals(1, $array['recipients']['to']);
+    }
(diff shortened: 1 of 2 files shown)


tests/Messages/BatchMessageTest.php
```diff
@@ -41,6 +41,20 @@ public function testAddRecipient()
...
+    {
+        $message = $this->client->BatchMessage($this->sampleDomain);
+        $message->addToRecipient('test@samples.mailgun.org', 'this-is-not-an-array');
+        $messageObj = $message->getMessage();
+        $this->assertEquals(['to' => ['test@samples.mailgun.org']], $messageObj);
+
+        $reflectionClass = new \ReflectionClass(get_class($message));
+        $property = $reflectionClass->getProperty('counters');
+        $property->setAccessible(true);
+        $array = $property->getValue($message);
```
Nyholm: What do you think of introducing (in the future) my NSA package? https://packagist.org/packages/nyholm/nsa

Nyholm:
Great. Thank you!

tests/Messages/BatchMessageTest.php
DavidGarciaCat: 👍 maybe we can use it when new v3 release is ready. For now I chose to use PHPUnit because all tests are using this package. Anyway, thanks for the idea!

Nyholm: No no, I do not want to get rid of PHPUnit. Just these four lines of code: 

```php
$reflectionClass = new \ReflectionClass(get_class($message));
$property = $reflectionClass->getProperty('counters');
$property->setAccessible(true);
$array = $property->getValue($message);
```
We have written that on several places. I suggest to use NSA because we can achieve the same with: 

```php
$array = NSA::getProperty($message, 'counters');
```

DavidGarciaCat: 👍 Understood! I know what you mean, I saw that piece of code too. Let's discuss how we can implement this on v3!

Issue #168, opened 2016-09-06
Retrieving stored messages


krystalcode:
Hello,

Does this library provide functionality for retrieving stored messages? I can get the url from the 'events' endpoint, but then do I have to use another http library such as curl to get the messages? It would be nice to use the same library for all endpoints.

krystalcode:
Here's some code I came up with for getting stored messages via events API, in case anybody else is trying to figure it out. Warning: this is just a test, not production ready.

```
$apiKey = 'your-api-key';
$domain = 'you-domain.com';

$mailgun = new Mailgun($apiKey);
$mailgun->setApiVersion('v3');
$eventsResponse = $mailgun->get("$domain/events", ['event' => 'stored']);
$responseBody = $eventsResponse->http_response_body;

if (!empty($responseBody->items)) {
  foreach ($responseBody->items as $item) {
    $url = $item->storage->url;
    $urlParts = explode('/', $url);

    $protocol    = array_shift($urlParts);
    $nothing     = array_shift($urlParts);
    $apiHost     = array_shift($urlParts);
    $apiVersion  = array_shift($urlParts);
    $endpointUrl = implode('/', $urlParts);

    $mailgun = new Mailgun($apiKey, null, $apiHost);
    $mailgun->setApiVersion($apiVersion);
    $response = $mailgun->get($endpointUrl);
    $responseBody = $response->http_response_body;
  }
}
```

It's a bit tedious though and would need even more boilerplate to scroll through paged results and using time ranges etc., so it would be nice to have the library provide some functionality that helps with all that, including Events API and retrieving individual messages via given urls.

Nyholm:
Hey. 
Thank you for this. I believe this should be done in a more easy way.

newscloud:
I opened this with Mailgun 23 days ago - theyre working on it ... here's copies of their replies. No word since 9/13.

Hi Jeff,

It looks like the issue here is that when using a GET request with the SDK it assumes the baseURL is going to be api.mailgun.net instead of so.api.mailgun.net or si.api.mailgun.net. 

I'm going to review this with the team to determine if this is the cause and will look into getting the PHP library updated with a fix.

I'll let you know once we have more information.

Thanks,
Chris@MG


Howdy Jeff,

Our apologies. We've been inundated with support request this week. Going to follow up with Chris and get you update. I don't have an ETA at this time to provide unfortunately.

Thanks,
Nolan
@Mailgun


Howdy Jeff,

[...]

PR #184, opened 2016-10-13
Access stored messages by URL

pirogoeth:
Addresses issues in #168

### DIFF
--- a/src/Mailgun/Connection/RestClient.php
+++ b/src/Mailgun/Connection/RestClient.php
@@ -349,6 +349,14 @@ class RestClient
     }
 
     /**
+     * @return string
+     */
+    public function getApiVersion()
+    {
+        return $this->apiVersion;
+    }
+
+    /**
      * @param bool $sslEnabled
      *
      * @return RestClient
--- a/src/Mailgun/Mailgun.php
+++ b/src/Mailgun/Mailgun.php
@@ -36,6 +36,19 @@ class Mailgun
     protected $apiKey;
 
     /**
+     * @var array [string => string]
+     */
+    private $regions = [
+        'si' => 'si.api.mailgun.net',
+        'so' => 'so.api.mailgun.net',
+    ];
+
+    /**
+     * @var array [string => RestClient]
+     */
+    private $regionClients = [];
+
+    /**
      * @param string|null $apiKey
      * @param HttpClient  $httpClient
      * @param string      $apiEndpoint
@@ -47,6 +60,51 @@ class Mailgun
     ) {
         $this->apiKey = $apiKey;
         $this->restClient = new RestClient($apiKey, $apiEndpoint, $httpClient);
+
+        $this->initializeRegionClients($httpClient);
+    }
+
+    /**
+     * This function creates clients for each individual region. Messages are
+     * only accessible from the region that received them, so having a mapping
+     * of region -> client seems reasonable.
+     *
+     * @param HttpClient|null $httpClient
+     *
+     * @return bool
+     */
+    private function initializeRegionClients(HttpClient $httpClient = null)
+    {
+        foreach ($this->regions as $region => $apiUrl) {
+            $this->regionClients[$region] = new RestClient(
+                $this->apiKey,
+                $apiUrl,
+                $httpClient
+            );
+        }
+
+        return true;
+    }
+
+    /**
+     * Tries to select the correct API region for a URL through simple
+     * comparisons with the known region URLs. If none are matched, null is
+     * returned.
+     *
+     * @param string $url
+     *
+     * @return \string
+     */
+    private function regionForUrl($url)
+    {
+        foreach ($this->regions as $region => $apiUrl) {
+            if (strpos($url, 'https://' . $apiUrl) === 0) {
+                return $region;
+            }
+        }
+
+        // Region in URL does *not* exist in the regions mapping.
+        return null;
     }
 
     /**
@@ -166,6 +224,14 @@ class Mailgun
     {
         $this->restClient->setApiVersion($apiVersion);
 
+        foreach ($this->regionClients as $regionClient) {
+            if ($regionClient === null) {
+                continue;
+            }
+
+            $regionClient->setApiVersion($apiVersion);
+        }
+
         return $this;
     }
 
@@ -207,5 +273,30 @@ class Mailgun
     public function BatchMessage($workingDomain, $autoSend = true)
     {
         return new BatchMessage($this->restClient, $workingDomain, $autoSend);
+    }
+
+    /**
+     * @param string $messageUrl
+     *
+     * @return \stdClass|null
+     */
+    public function getMessageFromUrl($messageUrl)
+    {
+        if ($messageUrl === null) {
+            return;
+        }
+
+        $reqPath = parse_url($messageUrl, PHP_URL_PATH);
+        $region = $this->regionForUrl($messageUrl);
+        $regionClient = $this->regionClients[$region];
+
+        // Use substr to trim off the API version path component.
+        $versionPrefix = "/{$regionClient->getApiVersion()}/";
+        $query = substr(
+            $reqPath,
+            strlen($versionPrefix)
+        );
+
+        return $regionClient->get($query, []);
     }
 }


src/Mailgun/Mailgun.php
```diff
@@ -36,20 +36,83 @@ class Mailgun
     protected $apiKey;
 
     /**
+     * @var [string => string]
+     */
+    protected $regions = [
```
Nyholm: This should be private.

src/Mailgun/Mailgun.php
```diff
@@ -36,20 +36,83 @@ class Mailgun
...
+        'si' => 'si.api.mailgun.net',
+        'so' => 'so.api.mailgun.net',
+    ];
+
+    /**
+     * @var [string => RestClient]
+     */
+    protected $regionClients = [];
```
Nyholm: This should be private.

src/Mailgun/Mailgun.php
```diff
@@ -36,20 +36,83 @@ class Mailgun
...
     /**
+     * @var [string => string]
+     */
+    protected $regions = [
+        'si' => 'si.api.mailgun.net',
+        'so' => 'so.api.mailgun.net',
+    ];
+
+    /**
+     * @var [string => RestClient]
```
Nyholm: `@var array [string => RestClient]`

Right? Or else PHPDoc will not pick this up. Same with $regions.

src/Mailgun/Mailgun.php
```diff
@@ -36,20 +36,83 @@ class Mailgun
...
      * @param HttpClient  $httpClient
      * @param string      $apiEndpoint
+     * @param bool        $noRegions
      */
     public function __construct(
         $apiKey = null,
         HttpClient $httpClient = null,
-        $apiEndpoint = 'api.mailgun.net'
+        $apiEndpoint = 'api.mailgun.net',
+        bool $noRegions = false
```
Nyholm: We are not exclusive for PHP7. Can't have `bool`.

It is generally bad practice to have feature flags like this.

src/Mailgun/Mailgun.php
```diff
@@ -36,20 +36,83 @@ class Mailgun
...
     ) {
         $this->apiKey = $apiKey;
         $this->restClient = new RestClient($apiKey, $apiEndpoint, $httpClient);
+
+        if ($noRegions !== true) {
+            $this->initializeRegionClients($httpClient);
```
Nyholm: This constructor would create 3 `RestClient` objects, right?

src/Mailgun/Mailgun.php
```diff
@@ -36,20 +36,83 @@ class Mailgun
...
+    {
+        foreach ($this->regions as $region => $apiUrl) {
+            if (strpos($url, 'https://' . $apiUrl) === 0) {
+                return $region;
+            }
+        }
     }
 
     /**
+     * Returns the client for the specified region.
```
Nyholm: This seams to be a loose comment.

src/Mailgun/Mailgun.php
```diff
@@ -208,4 +279,28 @@ public function BatchMessage($workingDomain, $autoSend = true)
     {
         return new BatchMessage($this->restClient, $workingDomain, $autoSend);
     }
+
+    /**
+     * @param string $messageUrl
+     *
+     * @return \stdClass
+     */
+    public function getMessageFromUrl($messageUrl = null)
```
Nyholm: Why would we allow this function to be called with no arguments?

src/Mailgun/Mailgun.php
```diff
@@ -208,4 +279,28 @@ public function BatchMessage($workingDomain, $autoSend = true)
...
+          return null;
+        }
+
+        $region = $this->regionForUrl($messageUrl);
+        $regionClient = $this->regionClients[$region];
+        $query = substr(
+            $messageUrl,
+            strspn(
+                $messageUrl,
+                'https://' . $this->regions[$region] . '/' . $regionClient->getApiVersion() . '/'
```
Nyholm: Wouldn't `'https://' . $this->regions[$region] . '/' . $regionClient->getApiVersion() . '/'` be equal to the first part of `$messageUrl` at this point?

Nyholm:
I do also miss unit tests and maybe function tests for this feature.

src/Mailgun/Mailgun.php
pirogoeth: Removed.

src/Mailgun/Mailgun.php
pirogoeth: Wow, missed that. Thanks!

src/Mailgun/Mailgun.php
pirogoeth: Good point. Fixed.

src/Mailgun/Connection/RestClient.php
```diff
@@ -349,6 +349,14 @@ public function setApiVersion($apiVersion)
     }
 
     /**
+     * @return string
+     */
+    public function getApiVersion()
```
pirogoeth: Let's talk about this tomorrow.

src/Mailgun/Mailgun.php
pirogoeth: Fixed.

src/Mailgun/Mailgun.php
pirogoeth: Fixed.

src/Mailgun/Mailgun.php
pirogoeth: Fixed.

pirogoeth:
@Nyholm Sorry about that, it's been a bit since I've used PHP. I'm gonna make sure everything gets fixed up and I'll have tests done either later today or tomorrow.

src/Mailgun/Mailgun.php
pirogoeth: Working on making this less crummy.

src/Mailgun/Connection/RestClient.php
DavidGarciaCat: From my POV I cannot see the problem to implement a getter. Given we already have a setter, a getter is not a bad idea. @Nyholm why do you think this is not a good idea?

src/Mailgun/Mailgun.php
```diff
@@ -166,6 +224,14 @@ public function setApiVersion($apiVersion)
     {
         $this->restClient->setApiVersion($apiVersion);
 
+        foreach ($this->regionClients as $regionClient) {
+            if ($regionClient == null) {
```
DavidGarciaCat: This should be compared strictly `===` or using the PHP function `is_null()`.

Also, please consider to use Yoda conditions, specially when you compare using `==`. Example:
`if ($regionClient = null)` will return **always** true and this is a human error but `if (null = $regionClient)` will return an error and it means we will prevent an unexpected behaviour.

src/Mailgun/Mailgun.php
```diff
@@ -208,4 +274,29 @@ public function BatchMessage($workingDomain, $autoSend = true)
...
+
+    /**
+     * @param string $messageUrl
+     *
+     * @return \stdClass
+     */
+    public function getMessageFromUrl($messageUrl)
+    {
+        if ($messageUrl === null) {
+            return null;
```
DavidGarciaCat: According to PSR standards, when we `return null;` we can just write `return;`

Also, when we return `null` is must be declared on `@return` annotation too, so `@return \stdClass|null`

src/Mailgun/Connection/RestClient.php
Nyholm: I would be happy to remove the setter as well =)

I have an idea that the RestClient should only care about building a request and handle the response. That includes adding headers and exception handling. 

To increase flexibility of the RestClient we should not allow it to know about API version or the URLs. That should be handled in the MailGun class.

DavidGarciaCat: That is a better approach, indeed. Good point.

src/Mailgun/Mailgun.php
Nyholm: Agreed. Even though `void !== null`

pirogoeth: Added null to the return annotation and dropped null from the return statement.

src/Mailgun/Mailgun.php
pirogoeth: Changed to strict comparison. I'll keep the Yoda conditions thing in mind, thanks! :)

src/Mailgun/Mailgun.php
```diff
@@ -47,6 +60,51 @@ public function __construct(
...
+    {
+        foreach ($this->regions as $region => $apiUrl) {
+            $this->regionClients[$region] = new RestClient(
+                $this->apiKey,
+                $apiUrl,
+                $httpClient
+            );
+        }
+
+        return true;
```
DavidGarciaCat: ❓ Question: Why to add here `return true;`? If we are not going to return any other value does not make sense - from my POV - to return always `true` so we can just remove this. Anyway, if we throw an Exception (for example) we are not going to reach that `return` so IMHO does not make sense or is not adding a useful case.

src/Mailgun/Mailgun.php
```diff
@@ -47,6 +60,51 @@ public function __construct(
...
+    private function regionForUrl($url)
+    {
+        foreach ($this->regions as $region => $apiUrl) {
+            if (strpos($url, 'https://' . $apiUrl) === 0) {
+                return $region;
+            }
+        }
+
+        // Region in URL does *not* exist in the regions mapping.
+        return null;
```
DavidGarciaCat: As discussed on my previous review, `return;` is the PSR standard for `return null;`

Also, I can see the `return @\string` annotation, but if I am not wrong `\string` is part of PHP7 annotations, isn't it? So I think a better approach here is `@return string` to be consistent with PHP5 code. In addition, given this method can return `null` too, it should be added to the annotation.

pirogoeth:
Closing this since it was a little premature on my part.

Issue #186, opened 2016-10-14
My thoughts for a new API


Nyholm:
Here are my thoughts for the completing the [Better API milestone](https://github.com/mailgun/mailgun-php/milestone/1). 

I believe we can make the api client much easier to work with and much easier to maintain. I’ve been looking at 20+ big API clients and there are basically 3 ways you can make a client. 

1. Pure HTTP: Does only provide authentication and a `client::api(string $url, array $params)`. Like [Happyr/LinkedIn](https://github.com/Happyr/LinkedIn-API-client)

2. Defined Resources: Defined all API resources and separate different parts of the API into API\Foo classes. Each resource of the API has a method in the API\Foo class. Parameters to these methods would be scalars and arrays. Like [KNP/Github](https://github.com/KnpLabs/php-github-api).

3. Domain models: Like *Defined Resources* but we also use domain objects for API responses and for POSTing data. 

Usability and the work to maintain the API increases from 1 to 3. 

The clear drawback of using *Domain models* is that we have to update the client every time the API updates. Whenever the API adds a parameter in the request or response we need to do an update that reflects that API change. That is not the case with *Defined resources*

I suggest we take a lot of inspiration from KNP/Github and do something similar ourselves. 

-----------------------

It is important that we can make this update in a backwards compatible way. I believe it is possible. We should make sure to deprecate functions and add more features one at the time. 

-----------------------

I suggest the following new classes: 
* [ ] API\Message
* [ ] API\Domain
* [ ] API\Event
* [x] API\Stats
* [ ] API\Tag
* [ ] API\Suppression
* [ ] API\Route
* [ ] API\Webhook (Possibly API\Domain\Webhook)
* [ ] API\List
* [ ] API\Address
* [ ] API\Campaign

These classes reflect the [API reference](https://documentation.mailgun.com/api_reference.html).

I also suggest we create a `Maingun::getUrl` to be able to fetch **any** url. We do only populate the Request with authentication headers. This will solve issues with fetching paginated data (#128) and stored messages (#168). 

------------------

### Return types
Today we return a \stdClass. I suggest we return an array response with the data we get from the server. 

We should throw exceptions on non-200 HTTP status. We should also throw an `MissingArgumentEception` or `InvalidArugmentException` if we notice that the arguments are invalid before we send the request. 

------------------

I would like some feedback on this change before we get started. Please give a :+1: or :-1: and maybe share a comment or two.

DavidGarciaCat:
[...]

PR #194, opened 2016-10-14
Added response classes

Nyholm:
This will address some issues in #186 

Is this something we want to do?
There is a lot of objects created. Should we consider performance here? Should we have a way to opt-out of this feature?


DavidGarciaCat:
❓ Question: What do you think about update the `AllResponse` class to `abstract class`? We can set a new class for each response type but keeping this process available, and - at the same time - we won't allow to transform from an Array just some fields instead of all of them.

Nyholm:
The name `AllResponse` is pretty bad. It should reflect the "Response" for the `Stats::all()` function. Just tile `TotalResponse` is for `Stats::total()`.

>  we won't allow to transform from an Array just some fields instead of all of them.

I do not understand. Can you elaborate?

DavidGarciaCat:
Sure,

What I am suggesting is to transform the `AllResponse` object to be an abstract like:
```
namespace Mailgun\Resource\Api\Stats;

abstract class AbstractResponse
{
    public static function createFromArray(array $array)
    {
        // some code here (shared fields)
        return $object;
    }
}
```

And then build a Response for each type, like:
```
namespace Mailgun\Resource\Api\Stats;

final class TotalResponse extends AbstractResponse
{
    public static function createFromArray(array $array)
    {
        $object = parent::createFromArray($array);
        // some code here (custom fields)
        return $object;
    }
}
```

Nyholm:
That would be good, but (as far as I can see) there is not shared fields. Just consider these two stats responses. They share nothing...

Nyholm:
I like this direction. Im going to merge it into my other PR.

Issue #187, opened 2016-10-14
Create Domains API


pirogoeth:
Includes request dispatcher and any request//response models necessary.

Nyholm:
Im going to close this as #187 is merged.

PR #198, opened 2016-10-19
Domains API, #187

pirogoeth:
This adds the `domains` API routes and the response objects that belong with it. Also fixes up some testing stuff, updates some requirements, and repairs some typos.

Still need:
 - [x] Integration tests
 - [x] More documentation (?)
 - [x] Snake case -> Camel case

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: php
- 
+
 php:
   - 5.5
   - 5.6
@@ -11,6 +11,9 @@ before_install:
 
 install:
   - travis_retry composer install
+  - travis_retry composer require --dev 'phpunit/php-code-coverage=^2.2.4'
 
 script:
-  - phpunit
+  - phpunit -c phpunit.xml.dist --testsuite unit --coverage-text
+  - phpunit -c phpunit.xml.dist --testsuite functional --coverage-text
+  - '[[ "${TRAVIS_PULL_REQUEST}" == "false" ]] && ( phpunit -c phpunit.xml.dist --testsuite integration --coverage-text ) || ( echo "Testing PR - No integration tests available")'
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         "webmozart/assert": "^1.1"
     },
     "require-dev": {
-        "phpunit/phpunit": "~4.6",
+        "phpunit/phpunit": "~4.8",
         "php-http/guzzle6-adapter": "^1.0"
     },
     "autoload": {
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -13,7 +13,26 @@
         <testsuite>
             <directory>tests</directory>
         </testsuite>
+
+        <testsuite name="unit">
+            <directory>tests</directory>
+            <exclude>tests/Integration</exclude>
+            <exclude>test/Functional</exclude>
+        </testsuite>
+
+        <testsuite name="integration">
+            <directory>tests/Integration</directory>
+        </testsuite>
+
+        <testsuite name="functional">
+            <directory>tests/Functional</directory>
+        </testsuite>
     </testsuites>
 
+    <filter>
+        <whitelist processUncoveredFilesFromWhitelist="true">
+            <directory suffix=".php">src</directory>
+        </whitelist>
+    </filter>
 
 </phpunit>
--- a/src/Mailgun/Api/AbstractApi.php
+++ b/src/Mailgun/Api/AbstractApi.php
@@ -4,10 +4,15 @@ namespace Mailgun\Api;
 
 use Http\Client\Common\HttpMethodsClient;
 use Http\Client\HttpClient;
-use Http\Message\RequestFactory;
-use Mailgun\Exception\HttpServerException;
 use Http\Client\Exception as HttplugException;
+use Http\Discovery\MessageFactoryDiscovery;
+use Http\Discovery\StreamFactoryDiscovery;
+use Http\Message\RequestFactory;
+use Http\Message\MultipartStream\MultipartStreamBuilder;
+use Mailgun\Assert;
+use Mailgun\Exception\HttpServerException;
 use Mailgun\Serializer\ResponseDeserializer;
+use Mailgun\Resource\Api\SimpleResponse;
 use Psr\Http\Message\ResponseInterface;
 
 /**
@@ -36,6 +41,25 @@ abstract class AbstractApi
     {
         $this->httpClient = new HttpMethodsClient($httpClient, $requestFactory);
         $this->serializer = $serializer;
+    }
+
+    /**
+     * Attempts to safely deserialize the response into the given class.
+     * If the HTTP return code != 200, deserializes into SimpleResponse::class
+     * to contain the error message and any other information provided.
+     *
+     * @param ResponseInterface $response
+     * @param string            $className
+     *
+     * @return $class|SimpleResponse
+     */
+    protected function safeDeserialize(ResponseInterface $response, $className)
+    {
+        if ($response->getStatusCode() !== 200) {
+            return $this->serializer->deserialize($response, SimpleResponse::class);
+        } else {
+            return $this->serializer->deserialize($response, $className);
+        }
     }
 
     /**
@@ -74,6 +98,20 @@ abstract class AbstractApi
     protected function post($path, array $parameters = [], array $requestHeaders = [])
     {
         return $this->postRaw($path, $this->createJsonBody($parameters), $requestHeaders);
+    }
+
+    /**
+     * Send a POST request with parameters encoded as multipart-stream form data.
+     *
+     * @param string $path           Request path.
+     * @param array  $parameters     POST parameters to be mutipart-stream-encoded.
+     * @param array  $requestHeaders Request headers.
+     *
+     * @return ResponseInterface
+     */
+    protected function postMultipart($path, array $parameters = [], array $requestHeaders = [])
+    {
+        return $this->doMultipart('POST', $path, $parameters, $requestHeaders);
     }
 
     /**
@@ -117,6 +155,20 @@ abstract class AbstractApi
     }
 
     /**
+     * Send a PUT request with parameters encoded as multipart-stream form data.
+     *
+     * @param string $path           Request path.
+     * @param array  $parameters     PUT parameters to be mutipart-stream-encoded.
+     * @param array  $requestHeaders Request headers.
+     *
+     * @return ResponseInterface
+     */
+    protected function putMultipart($path, array $parameters = [], array $requestHeaders = [])
+    {
+        return $this->doMultipart('PUT', $path, $parameters, $requestHeaders);
+    }
+
+    /**
      * Send a DELETE request with JSON-encoded parameters.
      *
      * @param string $path           Request path.
@@ -134,6 +186,60 @@ abstract class AbstractApi
         }
 
         return $response;
+    }
+
+    /**
+     * Send a DELETE request with parameters encoded as multipart-stream form data.
+     *
+     * @param string $path           Request path.
+     * @param array  $parameters     DELETE parameters to be mutipart-stream-encoded.
+     * @param array  $requestHeaders Request headers.
+     *
+     * @return ResponseInterface
+     */
+    protected function deleteMultipart($path, array $parameters = [], array $requestHeaders = [])
+    {
+        return $this->doMultipart('DELETE', $path, $parameters, $requestHeaders);
+    }
+
+    /**
+     * Send a request with parameters encoded as multipart-stream form data.
+     *
+     * @param string $type           Request type. (POST, PUT, etc.)
+     * @param string $path           Request path.
+     * @param array  $parameters     POST parameters to be mutipart-stream-encoded.
+     * @param array  $requestHeaders Request headers.
+     *
+     * @return ResponseInterface
+     */
+    protected function doMultipart($type, $path, array $parameters = [], array $requestHeaders = [])
+    {
+        Assert::oneOf(
+            $type,
+            [
+                'DELETE',
+                'POST',
+                'PUT',
+            ]
+        );
+
+        $streamFactory = StreamFactoryDiscovery::find();
+        $builder = new MultipartStreamBuilder($streamFactory);
+        foreach ($parameters as $k => $v) {
+            $builder->addResource($k, $v);
+        }
+
+        $multipartStream = $builder->build();
+        $boundary = $builder->getBoundary();
+
+        $request = MessageFactoryDiscovery::find()->createRequest(
+            $type,
+            $path,
+            ['Content-Type' => 'multipart/form-data; boundary='.$boundary],
+            $multipartStream
+        );
+
+        return $this->httpClient->sendRequest($request);
     }
 
     /**
--- a/src/Mailgun/Api/Domain.php
+++ b/src/Mailgun/Api/Domain.php
@@ -1,0 +1,269 @@
+<?php
+
+/*
+ * Copyright (C) 2013-2016 Mailgun
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+
+namespace Mailgun\Api;
+
+use Mailgun\Assert;
+use Mailgun\Resource\Api\Domain\ComplexDomain;
+use Mailgun\Resource\Api\Domain\Credential;
+use Mailgun\Resource\Api\Domain\CredentialListResponse;
+use Mailgun\Resource\Api\Domain\DeliverySettingsResponse;
+use Mailgun\Resource\Api\Domain\DeliverySettingsUpdateResponse;
+use Mailgun\Resource\Api\Domain\DomainListResponse;
+use Mailgun\Resource\Api\SimpleResponse;
+
+/**
+ * {@link https://documentation.mailgun.com/api-domains.html}.
+ *
+ * @author Sean Johnson <sean@mailgun.com>
+ */
+class Domain extends AbstractApi
+{
+    /**
+     * Returns a list of domains on the account.
+     *
+     * @param int $limit
+     * @param int $skip
+     *
+     * @return DomainListResponse
+     */
+    public function listAll($limit = 100, $skip = 0)
+    {
+        Assert::integer($limit);
+        Assert::integer($skip);
+
+        $params = [
+            'limit' => $limit,
+            'skip' => $skip,
+        ];
+
+        $response = $this->get('/v3/domains', $params);
+
+        return $this->serializer->deserialize($response, DomainListResponse::class);
+    }
+
+    /**
+     * Returns a single domain.
+     *
+     * @param string $domain Name of the domain.
+     *
+     * @return ComplexDomain|array|ResponseInterface
+     */
+    public function info($domain)
+    {
+        Assert::stringNotEmpty($domain);
+
+        $response = $this->get(sprintf('/v3/domains/%s', $domain));
+
+        return $this->serializer->deserialize($response, ComplexDomain::class);
+    }
+
+    /**
+     * Creates a new domain for the account.
+     * See below for spam filtering parameter information.
+     * {@link https://documentation.mailgun.com/user_manual.html#um-spam-filter}.
+     *
+     * @param string $domain     Name of the domain.
+     * @param string $smtpPass   Password for SMTP authentication.
+     * @param string $spamAction `disable` or `tag` - inbound spam filtering.
+     * @param bool   $wildcard   Domain will accept email for subdomains.
+     *
+     * @return ComplexDomain|array|ResponseInterface
+     */
+    public function create($domain, $smtpPass, $spamAction, $wildcard)
+    {
+        Assert::stringNotEmpty($domain);
+        Assert::stringNotEmpty($smtpPass);
+        // TODO(sean.johnson): Extended spam filter input validation.
+        Assert::stringNotEmpty($spamAction);
+        Assert::boolean($wildcard);
+
+        $params = [
+            'name' => $domain,
+            'smtp_password' => $smtpPass,
+            'spam_action' => $spamAction,
+            'wildcard' => $wildcard,
+        ];
+
+        $response = $this->postMultipart('/v3/domains', $params);
+
+        return $this->safeDeserialize($response, ComplexDomain::class);
+    }
+
+    /**
+     * Removes a domain from the account.
+     * WARNING: This action is irreversible! Be cautious!
+     *
+     * @param string $domain Name of the domain.
+     *
+     * @return SimpleResponse|array|ResponseInterface
+     */
+    public function remove($domain)
+    {
+        Assert::stringNotEmpty($domain);
+
+        $response = $this->delete(sprintf('/v3/domains/%s', $domain));
+
+        return $this->serializer->deserialize($response, SimpleResponse::class);
+    }
+
+    /**
+     * Returns a list of SMTP credentials for the specified domain.
+     *
+     * @param string $domain Name of the domain.
+     * @param int    $limit  Number of credentials to return
+     * @param int    $skip   Number of credentials to omit from the list
+     *
+     * @return CredentialsListResponse
+     */
+    public function listCredentials($domain, $limit = 100, $skip = 0)
+    {
+        Assert::stringNotEmpty($domain);
+        Assert::integer($limit);
+        Assert::integer($skip);
+
+        $params = [
+            'limit' => $limit,
+            'skip' => $skip,
+        ];
+
+        $response = $this->get(sprintf('/v3/domains/%s/credentials', $domain), $params);
+
+        return $this->safeDeserialize($response, CredentialListResponse::class);
+    }
+
+    /**
+     * Create a new SMTP credential pair for the specified domain.
+     *
+     * @param string $domain   Name of the domain.
+     * @param string $login    SMTP Username.
+     * @param string $password SMTP Password. Length min 5, max 32.
+     *
+     * @return Credential|array|ResponseInterface
+     */
+    public function newCredential($domain, $login, $password)
+    {
+        Assert::stringNotEmpty($domain);
+        Assert::stringNotEmpty($login);
+        Assert::stringNotEmpty($password);
+        Assert::lengthBetween($password, 5, 32, 'SMTP password must be between 5 and 32 characters.');
+
+        $params = [
+            'login' => $login,
+            'password' => $password,
+        ];
+
+        $response = $this->postMultipart(sprintf('/v3/domains/%s/credentials', $domain), $params);
+
+        return $this->serializer->deserialize($response, SimpleResponse::class);
+    }
+
+    /**
+     * Update a set of SMTP credentials for the specified domain.
+     *
+     * @param string $domain Name of the domain.
+     * @param string $login  SMTP Username.
+     * @param string $pass   New SMTP Password. Length min 5, max 32.
+     *
+     * @return SimpleResponse|array|ResponseInterface
+     */
+    public function updateCredential($domain, $login, $pass)
+    {
+        Assert::stringNotEmpty($domain);
+        Assert::stringNotEmpty($login);
+        Assert::stringNotEmpty($pass);
+        Assert::lengthBetween($pass, 5, 32, 'SMTP password must be between 5 and 32 characters.');
+
+        $params = [
+            'password' => $pass,
+        ];
+
+        $response = $this->putMultipart(
+            sprintf(
+                '/v3/domains/%s/credentials/%s',
+                $domain,
+                $login
+            ),
+            $params
+        );
+
+        return $this->serializer->deserialize($response, SimpleResponse::class);
+    }
+
+    /**
+     * Remove a set of SMTP credentials from the specified domain.
+     *
+     * @param string $domain Name of the domain.
+     * @param string $login  SMTP Username.
+     *
+     * @return SimpleResponse|array|ResponseInterface
+     */
+    public function deleteCredential($domain, $login)
+    {
+        Assert::stringNotEmpty($domain);
+        Assert::stringNotEmpty($login);
+
+        $response = $this->delete(
+            sprintf(
+                '/v3/domains/%s/credentials/%s',
+                $domain,
+                $login
+            )
+        );
+
+        return $this->serializer->deserialize($response, SimpleResponse::class);
+    }
+
+    /**
+     * Returns delivery connection settings for the specified domain.
+     *
+     * @param string $domain Name of the domain.
+     *
+     * @return DeliverySettingsResponse|array|ResponseInterface
+     */
+    public function getDeliverySettings($domain)
+    {
+        Assert::stringNotEmpty($domain);
+
+        $response = $this->get(sprintf('/v3/domains/%s/connection', $domain));
+
+        return $this->serializer->deserialize($response, DeliverySettingsResponse::class);
+    }
+
+    /**
+     * Updates the specified delivery connection settings for the specified domain.
+     * If a parameter is passed in as null, it will not be updated.
+     *
+     * @param string    $domain     Name of the domain.
+     * @param bool|null $requireTLS Enforces that messages are sent only over a TLS connection.
+     * @param bool|null $noVerify   Disables TLS certificate and hostname verification.
+     *
+     * @return DeliverySettingsResponse|array|ResponseInterface
+     */
+    public function updateDeliverySettings($domain, $requireTLS, $noVerify)
+    {
+        Assert::stringNotEmpty($domain);
+        Assert::nullOrBoolean($requireTLS);
+        Assert::nullOrBoolean($noVerify);
+
+        $params = [];
+
+        if (null !== $requireTLS) {
+            $params['require_tls'] = $requireTLS ? 'true' : 'false';
+        }
+
+        if (null !== $noVerify) {
+            $params['skip_verification'] = $noVerify ? 'true' : 'false';
+        }
+
+        $response = $this->putMultipart(sprintf('/v3/domains/%s/connection', $domain), $params);
+
+        return $this->serializer->deserialize($response, DeliverySettingsUpdateResponse::class);
+    }
+}
--- a/src/Mailgun/Mailgun.php
+++ b/src/Mailgun/Mailgun.php
@@ -260,4 +260,12 @@ class Mailgun
     {
         return new Api\Stats($this->httpClient, $this->requestFactory, $this->serializer);
     }
+
+    /**
+     * @return Api\Domain
+     */
+    public function getDomainApi()
+    {
+        return new Api\Domain($this->httpClient, $this->requestFactory, $this->serializer);
+    }
 }
--- a/src/Mailgun/Resource/Api/Domain/ComplexDomain.php
+++ b/src/Mailgun/Resource/Api/Domain/ComplexDomain.php
@@ -1,0 +1,95 @@
+<?php
+
+/**
+ * Copyright (C) 2013-2016 Mailgun.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+namespace Mailgun\Resource\Api\Domain;
+
+use Mailgun\Assert;
+use Mailgun\Resource\CreatableFromArray;
+
+/**
+ * ComplexDomain uses DomainTrait and exposes a "complex" constructor
+ * where an array or \stdClass can be passed in to find the appropriate
+ * fields.
+ *
+ * @author Sean Johnson <sean@mailgun.com>
+ */
+class ComplexDomain implements CreatableFromArray
+{
+    /**
+     * @var SimpleDomain
+     */
+    private $domain;
+
+    /**
+     * @var DomainDnsRecord[]
+     */
+    private $inboundDnsRecords;
+
+    /**
+     * @var DomainDnsRecord[]
+     */
+    private $outboundDnsRecords;
+
+    /**
+     * @param array $data
+     *
+     * @return ComplexDomain
+     */
+    public static function createFromArray(array $data)
+    {
+        Assert::keyExists($data, 'domain');
+        Assert::keyExists($data, 'receiving_dns_records');
+        Assert::keyExists($data, 'sending_dns_records');
+
+        // Let DomainDnsRecord::createFromArray() handle validation of
+        // the `receiving_dns_records` and `sending_dns_records` data.
+        // Also let SimpleDomain::createFromArray() handle validation of
+        // the `domain` fields.
+        return new static(
+            SimpleDomain::createFromArray($data['domain']),
+            DomainDnsRecord::createFromArray($data['receiving_dns_records']),
+            DomainDnsRecord::createFromArray($data['sending_dns_records'])
+        );
+    }
+
+    /**
+     * @param SimpleDomain $domainInfo
+     * @param array        $rxRecords  Array of DomainDnsRecord instances
+     * @param array        $txRecords  Array of DomainDnsRecord instances
+     */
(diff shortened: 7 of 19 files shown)


.travis.yml
```diff
@@ -13,4 +13,4 @@ install:
   - travis_retry composer install
 
 script:
-  - phpunit
\ No newline at end of file
+  - phpunit -c phpunit.xml --tap --coverage-text
```
Nyholm: There is no (and should not be) - phpunit.xml
Do we really want to run test coverage on all commits?

Have a look at what we do here: https://github.com/php-http/message/blob/master/.travis.yml

That is a good solution. But make that a separate PR.

composer.json
```diff
@@ -11,8 +11,9 @@
         "webmozart/assert": "^1.2@dev"
     },
     "require-dev": {
-        "phpunit/phpunit": "~4.6",
-        "php-http/guzzle6-adapter": "^1.0"
+        "phpunit/phpunit": "~5.6",
```
Nyholm: This is not compatible with php 5.5

phpunit.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Nyholm: This file should be removed. Prefer adding it to gitignore

modd.conf
```diff
@@ -0,0 +1,4 @@
+tests/** {
```
Nyholm: Why do we need this file?

phpunit.xml.dist
```diff
@@ -1,19 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
Nyholm: We should keep the dist file. If anyone want to change the settings they will copy phpunit.xml.dist to phpunit.xml. But the latter file should not be in git.

src/Mailgun/Api/Domains.php
```diff
@@ -0,0 +1,270 @@
...
+use Mailgun\Resource\Api\Domains\DomainListResponse;
+use Mailgun\Resource\Api\Domains\SettingsUpdateResponse;
+use Mailgun\Resource\Api\SimpleResponse;
+
+/**
+ * {@link https://documentation.mailgun.com/api-domains.html}
+ *
+ * @author Sean Johnson <sean@mailgun.com>
+ */
+class Domains extends AbstractApi
```
Nyholm: Classes and namespaces should be in singular.

src/Mailgun/Api/Domains.php
```diff
@@ -0,0 +1,270 @@
...
+        return $this->serializer->deserialize($response, ComplexDomain::class);
+    }
+
+    /**
+     * Removes a domain from the account.
+     * WARNING: This action is irreversible! Be cautious!
+     *
+     * @param string    $domain     Name of the domain.
+     *
+     * @return ...
```
Nyholm: We need a return type.

src/Mailgun/Api/Domains.php
```diff
@@ -0,0 +1,270 @@
...
+     * Creates a new domain for the account.
+     * See below for spam filtering parameter information.
+     * {@link https://documentation.mailgun.com/user_manual.html#um-spam-filter}
+     *
+     * @param string    $domain     Name of the domain.
+     * @param string    $smtpPass   Password for SMTP authentication.
+     * @param string    $spamAction `disable` or `tag` - inbound spam filtering.
+     * @param bool      $wildcard   Domain will accept email for subdomains.
+     *
+     * @return ComplexDomain
```
Nyholm: ComplexDomain|array|ResponseInterface

Same for all

src/Mailgun/Api/Domains.php
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    /**
+     * Returns a list of SMTP credentials for the specified domain.
+     *
+     * @param string    $domain     Name of the domain.
+     *
+     * @return CredentialsListResponse
+     */
+    public function credentials_list($domain, $limit = 100, $skip = 0)
```
Nyholm: One should never use snake_case

src/Mailgun/Api/Domains.php
```diff
@@ -0,0 +1,270 @@
...
+    /**
+     * Create a new SMTP credential pair for the specified domain.
+     *
+     * @param string    $domain     Name of the domain.
+     * @param string    $login      SMTP Username.
+     * @param string    $pass       SMTP Password. Length min 5, max 32.
+     *
+     * @return CredentialResponse
+     */
+    public function new_credential($domain, $login, $pass)
```
Nyholm: Same here. Use newCredential

src/Mailgun/Api/Domains.php
```diff
@@ -0,0 +1,270 @@
...
+    {
+        Assert::stringNotEmpty($domain);
+        Assert::stringNotEmpty($login);
+        Assert::stringNotEmpty($pass);
+        Assert::lengthBetween($pass, 5, 32, 'SMTP password must be between 5 and 32 characters.');
```
Nyholm: Good!

src/Mailgun/Api/Domains.php
```diff
@@ -0,0 +1,270 @@
...
+                $login
+            ),
+            $params
+        );
+
+        return $this->serializer->deserialize($response, SimpleResponse::class);
+    }
+
+    /**
+     * Returns delivery connection settings for the specified domain.
```
Nyholm: Really good that you document them well

src/Mailgun/Api/Stats.php
```diff
@@ -25,7 +25,7 @@ public function total($domain, array $params = [])
 
         $response = $this->get(sprintf('/v3/%s/stats/total', rawurlencode($domain)), $params);
 
-        return $this->serializer->deserialze($response, TotalResponse::class);
+        return $this->serializer->deserialize($response, TotalResponse::class);
```
Nyholm: Oops. Thank you

src/Mailgun/Mailgun.php
```diff
@@ -108,6 +107,10 @@ public function api($name)
             case 'stats':
                 $api = new Api\Stats($this->httpClient, $this->requestFactory, $this->serializer);
                 break;
+            case 'domains':
+            case 'domain':
```
Nyholm: I'm okey with this. But remember to att the method annotation to the class

src/Mailgun/Resource/Api/Domains/ComplexDomain.php
```diff
@@ -0,0 +1,81 @@
...
+namespace Mailgun\Resource\Api\Domains;
+
+/**
+ * ComplexDomain uses DomainTrait and exposes a "complex" constructor
+ * where an array or \stdClass can be passed in to find the appropriate
+ * fields.
+ *
+ * @author Sean Johnson <sean@mailgun.com>
+ */
+class ComplexDomain
```
Nyholm: All of these should implement the CreatableFromData interface.

src/Mailgun/Resource/Api/SimpleResponse.php
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Only set when API rate limit is hit and a rate limit response is returned.
+     *
+     * @var int
+     */
+    private $retrySeconds = null;
+
+    /**
+     * Only set on calls such as DELETE /v3/domains/.../credentials/<user>.
```
Nyholm: What is $spec?

src/Mailgun/Resource/Api/SimpleResponse.php
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+namespace Mailgun\Resource\Api;
+
+/**
+ * @author Sean Johnson <sean@mailgun.com>
+ */
+class SimpleResponse
```
Nyholm: It sure about this class. But I understand the need for it...

Does it miss httpResponseCode?

src/Mailgun/Resource/Api/Domains/SimpleDomain.php
```diff
@@ -0,0 +1,128 @@
+<?php
+
+/**
+ * Copyright (C) 2013-2016 Mailgun.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license.  See the LICENSE file for details.
+ */
+namespace Mailgun\Resource\Api\Domains;
```
Nyholm: Namespace should be singular.

tests/Api/DomainsTest.php
```diff
@@ -0,0 +1,62 @@
...
+ */
+class DomainsTest extends TestCase
+{
+    protected function getApiClass()
+    {
+        return 'Mailgun\Api\Domains';
+    }
+
+    /**
+     * NOTE: This test uses the production Mailgun API.
```
Nyholm: Why? This should be avoided. If we need it we should group these tests and exclude them by default.

src/Mailgun/Api/Domains.php
```diff
@@ -0,0 +1,270 @@
...
+
+        $response = $this->get('/v3/domains', $params);
+
+        return $this->serializer->deserialize($response, DomainListResponse::class);
+    }
+
+    /**
+     * Returns a single domain.
+     *
+     * @param string    $domain     Name of the domain.
```
DavidGarciaCat: There are no other `@param` to this should be:
`@param string $domain Name of the domain.`

src/Mailgun/Api/Domains.php
```diff
@@ -0,0 +1,270 @@
...
+        $response = $this->post('/v3/domains', $params);
+
+        return $this->serializer->deserialize($response, ComplexDomain::class);
+    }
+
+    /**
+     * Removes a domain from the account.
+     * WARNING: This action is irreversible! Be cautious!
+     *
+     * @param string    $domain     Name of the domain.
```
DavidGarciaCat: Same case, already described above.

src/Mailgun/Api/Domains.php
```diff
@@ -0,0 +1,270 @@
...
+
+        $response = $this->delete(sprintf('/v3/domains/%s', $domain));
+
+        return $this->serializer->deserialize($response, SimpleResponse::class);
+    }
+
+    /**
+     * Returns a list of SMTP credentials for the specified domain.
+     *
+     * @param string    $domain     Name of the domain.
```
DavidGarciaCat: And once more here 😅

DavidGarciaCat: Missed `@param` annotations:
```
* @param int $limit
* @param int $skip
```

src/Mailgun/Api/Domains.php
```diff
@@ -0,0 +1,270 @@
...
+
+        return $this->serializer->deserialize($response, CredentialListResponse::class);
+    }
+
+    /**
+     * Create a new SMTP credential pair for the specified domain.
+     *
+     * @param string    $domain     Name of the domain.
+     * @param string    $login      SMTP Username.
+     * @param string    $pass       SMTP Password. Length min 5, max 32.
```
DavidGarciaCat:
```
* @param string $domain Name of the domain.
* @param string $login  SMTP Username.
* @param string $pass   SMTP Password. Length min 5, max 32.
```

src/Mailgun/Api/Domains.php
DavidGarciaCat: I know that this is the API requirement (`A password for the SMTP credentials. (Length Min 5, Max 32)`) but personally I will never understand why some environments are setting a Max Length on password fields. Does anyone know why is there this "limitation"?

src/Mailgun/Api/Domains.php
```diff
@@ -0,0 +1,270 @@
...
+
+        return $this->serializer->deserialize($response, CredentialResponse::class);
+    }
+
+    /**
+     * Update a set of SMTP credentials for the specified domain.
+     *
+     * @param string    $domain     Name of the domain.
+     * @param string    $login      SMTP Username.
+     * @param string    $pass       New SMTP Password. Length min 5, max 32.
```
DavidGarciaCat: Same case here

src/Mailgun/Api/Domains.php
```diff
@@ -0,0 +1,270 @@
...
+     *
+     * @return CredentialResponse
+     */
+    public function update_credential($domain, $login, $pass)
```
DavidGarciaCat: snake_case here

src/Mailgun/Api/Domains.php
```diff
@@ -0,0 +1,270 @@
...
+        );
+
+        return $this->serializer->deserialize($response, SimpleResponse::class);
+    }
+
+    /**
+     * Remove a set of SMTP credentials from the specified domain.
+     *
+     * @param string    $domain     Name of the domain.
+     * @param string    $login      SMTP Username.
```
DavidGarciaCat: Same case here

src/Mailgun/Api/Domains.php
```diff
@@ -0,0 +1,270 @@
...
+     *
+     * @return ...
```
DavidGarciaCat: `@return` should have a type hint

src/Mailgun/Api/Domains.php
```diff
@@ -0,0 +1,270 @@
...
+     */
+    public function delete_credential($domain, $login)
```
DavidGarciaCat: snake_case

src/Mailgun/Api/Domains.php
```diff
@@ -0,0 +1,270 @@
...
+            $params
+        );
+
+        return $this->serializer->deserialize($response, SimpleResponse::class);
+    }
+
+    /**
+     * Returns delivery connection settings for the specified domain.
+     *
+     * @param string    $domain     Name of the domain.
```
DavidGarciaCat: Same case here

src/Mailgun/Api/Domains.php
```diff
@@ -0,0 +1,270 @@
...
+     *
+     * @return DeliverySettingsResponse
+     */
+    public function get_delivery_settings($domain)
```
DavidGarciaCat: snake_case

src/Mailgun/Api/Domains.php
```diff
@@ -0,0 +1,270 @@
...
+     * Updates the specified delivery connection settings for the specified domain.
+     * If a parameter is passed in as null, it will not be updated.
+     *
+     * @param string    $domain     Name of the domain.
+     * @param bool|null $requireTLS Enforces that messages are sent only over a TLS connection.
+     * @param bool|null $noVerify   Disables TLS certificate and hostname verification.
+     *
+     * @return DeliverySettingsResponse
+     */
+    public function update_delivery_settings($domain, $requireTLS, $noVerify)
```
DavidGarciaCat: snake_case

src/Mailgun/Mailgun.php
DavidGarciaCat: Why to set `domain`? The API Docs reference contains `domains` but not `domain`

src/Mailgun/Resource/Api/Domains/Credential.php
```diff
@@ -0,0 +1,75 @@
...
+     * @var string
+     */
+    private $mailbox;
+
+    /**
+     * @var string
+     */
+    private $login;
+
+    public function __construct($sizeBytes, \DateTime $createdAt, $mailbox, $login)
```
DavidGarciaCat: Missed annotation here

src/Mailgun/Resource/Api/Domains/CredentialListResponse.php
```diff
@@ -0,0 +1,75 @@
...
+    public static function createFromArray(array $data)
+    {
+        $items = [];
+
+        foreach ($data['items'] as $item) {
+            $items[] = new Credential(
+                $item['size_bytes'],
+                new \DateTime($item['created_at']),
+                $item['mailbox'],
+                $item['login']
```
DavidGarciaCat: Personally I like to check if the array has the expected key.

So maybe you can check - at least - if there is the `array_key_exists('items', $data)`. Also, given this is an array, provided as argument and we need to use these keys in order to build the `Credential` object, may be a good idea to check if the `$item` has all expected keys.

Is up to you what we need to do here: throw an exception or just ignore the invalid `$item` (if it does not contain all expected keys)

src/Mailgun/Resource/Api/Domains/CredentialListResponse.php
```diff
@@ -0,0 +1,75 @@
...
+            );
+        }
+
+        return new self($data['total_count'], $items);
```
DavidGarciaCat: Same case here: `$data['total_count']`

src/Mailgun/Resource/Api/Domains/DeliverySettingsResponse.php
```diff
@@ -0,0 +1,51 @@
...
+     * @var string
+     */
+    private $noVerify;
+
+    /**
+     * @var string
+     */
+    private $requireTLS;
+
+    public function __construct($data)
```
DavidGarciaCat: Missed annotation

src/Mailgun/Resource/Api/Domains/DeliverySettingsResponse.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+        $connSettings = $data['connection'];
+        $this->noVerify = $connSettings['skip_verification'] ? 'true' : 'false';
+        $this->requireTLS = $connSettings['require_tls'] ? 'true' : 'false';
```
DavidGarciaCat: You're assuming the expected key (`$data['connection']`) exists without validation. Then you are just checking if a sub-key exists to set a predefined value.

Instead of this, I would like to suggest to set 2 arguments on this constructor, and set the values directly (don't forget to validate them, to ensure we get valid values).

src/Mailgun/Resource/Api/Domains/DomainDnsRecord.php
```diff
@@ -0,0 +1,130 @@
...
+    {
+        $items = [];
+
+        foreach ($data as $item) {
+            $items[] = new static(
+                array_key_exists('name', $item) ? $item['name'] : null,
+                $item['record_type'],
+                $item['value'],
+                array_key_exists('priority', $item) ? $item['priority'] : null,
+                $item['valid']
```
DavidGarciaCat: Again assuming all keys exist

src/Mailgun/Resource/Api/Domains/DomainListResponse.php
```diff
@@ -0,0 +1,78 @@
...
+
+        foreach ($data['items'] as $item) {
+            $items[] = new SimpleDomain(
+                $item['name'],
+                $item['smtp_login'],
+                $item['smtp_password'],
+                $item['wildcard'],
+                $item['spam_action'],
+                $item['state'],
+                new \DateTime($item['created_at'])
```
DavidGarciaCat: And once more here there is no validation 😃

src/Mailgun/Resource/Api/Domains/DomainResponse.php
```diff
@@ -0,0 +1,37 @@
...
+
+        foreach ($data['items'] as $item) {
+            $items[] = new Domain(
+                $item['name'],
+                $item['smtp_login'],
+                $item['smtp_password'],
+                $item['wildcard'],
+                $item['spam_action'],
+                $item['state'],
+                new \DateTime($item['created_at'])
```
DavidGarciaCat: Same case

src/Mailgun/Resource/Api/Domains/SettingsUpdateResponse.php
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * @var string
+     */
+    private $requireTLS;
+
+    public function __construct($data)
+    {
+        $this->message = $data['message'];
+        $this->noVerify = $data['skip_verification'] ? 'true' : 'false';
+        $this->requireTLS = $data['require_tls'] ? 'true' : 'false';
```
DavidGarciaCat: Same case

src/Mailgun/Resource/Api/SimpleResponse.php
```diff
@@ -0,0 +1,73 @@
...
+    private $retrySeconds = null;
+
+    /**
+     * Only set on calls such as DELETE /v3/domains/.../credentials/<user>.
+     *
+     * @var string
+     */
+    private $spec = null;
+
+    public function __construct(array $data)
```
DavidGarciaCat: Missed annotation

src/Mailgun/Resource/Api/SimpleResponse.php
```diff
@@ -0,0 +1,73 @@
...
+    {
+        $this->message = $data['message'];
```
DavidGarciaCat: Missed validation

tests/Api/DomainsTest.php
pirogoeth: But integration testing..? I have no problem with have a separate group for integration tests vs. functional, etc.

.travis.yml
pirogoeth: I will make this a separate PR.

Why should there not be a phpunit config for our preferred testing settings, especially for CI? Also, why would we not want to run coverage on all commits? I feel that constantly monitoring test coverage is important, especially while the project is under active restructuring like this.

composer.json
pirogoeth: Ah, did not know this. I'll look into/downgrade this tomorrow. :\

src/Mailgun/Api/Domains.php
pirogoeth: :+1:

src/Mailgun/Api/Domains.php
pirogoeth: Yeah, still a work in progress.

.travis.yml
Nyholm: Sorry, I did not mean commits. I meant "test run". If we test of 5 different PHP versions, why should we run the coverage on all of them? The coverage will be the same, right? It takes simnifically more time to run the tests if you want to calculate the coverage. That is why you only run the coverage once per CI run.

src/Mailgun/Resource/Api/SimpleResponse.php
pirogoeth: The API mentions the existence of a spec field for some generic, simple responses, but there is not description / example of what it is. I will look into this tomorrow.

tests/Api/DomainsTest.php
Nyholm: Sure, integration testing is important. But it should not be mixed with unit testing or functional testing.

src/Mailgun/Resource/Api/SimpleResponse.php
pirogoeth: No, this is most literally the simplest response the Mailgun API will return. There are quite a few calls that return a simple response, such as {"message": "..."} or a rate-limit response {"retry-seconds": ...}. We could attempt to add an `httpResponseCode` member, but we would either need to have a common abstraction for all of the classes that are a "response" that also grabs / handles the `httpResponseCode` or handle it in some other way.

.travis.yml
pirogoeth: I agree with this. We can certainly set up a single step for running coverage only once. :+1:

pirogoeth:
@DavidGarciaCat Thanks for all the style feedback - I applied a StyleCI fix PR to my branch and I'll finish with the data validation, renaming, etc., tomorrow!

@Nyholm I'll migrate some stuff around and have a specific subdirectory in the tests for integration tests, and once we've got that we can set up the groups to keep things separate during the test run.

Appreciate the feedback, guys!

src/Mailgun/Api/Domains.php
pirogoeth: :+1:

src/Mailgun/Api/Domains.php
pirogoeth: :+1:

src/Mailgun/Resource/Api/Domains/ComplexDomain.php
pirogoeth: :+1:

src/Mailgun/Resource/Api/Domains/SimpleDomain.php
pirogoeth: :+1:

src/Mailgun/Mailgun.php
Nyholm: Yeah. I've just updated my PR

src/Mailgun/Api/Domains.php
pirogoeth: @DavidGarciaCat so I spent a good amount of time looking in to this... and as far as I can tell, there is not really a reason. I am going to be looking in to it some more shortly.

.travis.yml
Nyholm: It should. But that file is called phpunit.xml.dist

src/Mailgun/Resource/Api/Domains/CredentialListResponse.php
Nyholm: Interesting question you bring upp here. Should we trust the API response or not?

Should we blindly assume the API does as the doc says or should we verify that at each request?

src/Mailgun/Resource/Api/SimpleResponse.php
pirogoeth: In the case of `DELETE /domains/<domain>/credentials/<login>`, $spec is the login name/mailbox name of the credential pair that was deleted.

src/Mailgun/Mailgun.php
pirogoeth: Okay.

Nyholm:
I wrote a comment: https://github.com/mailgun/mailgun-php/pull/198#discussion_r84571300

> Interesting question you bring upp here. Should we trust the API response or not?

> Should we blindly assume the API does as the doc says or should we verify that at each request?

What do you think?

pirogoeth:
@Nyholm We should verify each response. I know we are also validating user input, but there are still cases where the API won't return what it says it will return. For example, adding a duplicate domain will cause the API to return only a `message` field in the response. I've slightly mitigated this with `AbstractApi::safeDeserialize()` but I still think it would be best to verify each API response.

pirogoeth:
Also - Travis doesn't expose envvars to PRs during the run. Here is a link to Travis with a build of my fork running integration tests. https://travis-ci.org/pirogoeth/mailgun-php/builds/170270900

Nyholm:
> @Nyholm We should verify each response. I know we are also validating user input, but there are still cases where the API won't return what it says it will return. For example, adding a duplicate domain will cause the API to return only a message field in the response. I've slightly mitigated this with AbstractApi::safeDeserialize() but I still think it would be best to verify each API response.

If the validation fails, should we throw a ServerException or a ClientException? Right now we throw an InvalidArgument exception which is wrong because the user did not use any invalid arguments. 

I do not think validation is needed. If the documentation differs from what is actually returned then we should update the documentation. If the server sometimes returns something unexpected, then it is an error on the API server that should be fixed.

.travis.yml
```diff
@@ -13,4 +13,5 @@ install:
   - travis_retry composer install
 
 script:
-  - phpunit
\ No newline at end of file
+  - phpunit --exclude-group integration
```
Nyholm: Put the exclude in phpunit.xml.dist

.travis.yml
```diff
@@ -13,4 +13,5 @@ install:
...
+  - '[[ "${TRAVIS_PULL_REQUEST}" == "false" ]] && ( phpunit --group integration ) || ( echo "Testing PR - No integration tests available")'
```
Nyholm: Why `"${TRAVIS_PULL_REQUEST}" == "false"`?

Is it to reduce the number of request to the server? If so, then it is okey (but weird). 

If it is to make sure we do not allow others for sending emails by making a PR? Then it's wrong. I'll just remove this line in by PR.

composer.json
```diff
@@ -11,8 +11,9 @@
         "webmozart/assert": "^1.1"
     },
     "require-dev": {
-        "phpunit/phpunit": "~4.6",
-        "php-http/guzzle6-adapter": "^1.0"
+        "phpunit/phpunit": "~4.8",
+        "php-http/guzzle6-adapter": "^1.0",
+        "phpunit/php-code-coverage": "^2.2.4"
```
Nyholm: Is this needed as a dependency?
Couldn't this be added on travis only?

src/Mailgun/Api/AbstractApi.php
```diff
@@ -39,6 +44,25 @@ public function __construct(HttpClient $httpClient, RequestFactory $requestFacto
...
+     * Attempts to safely deserialize the response into the given class.
+     * If the HTTP return code != 200, deserializes into SimpleResponse::class
+     * to contain the error message and any other information provided.
+     *
+     * @param ResponseInterface $response
+     * @param string            $className
+     *
+     * @return $class|SimpleResponse
+     */
+    protected function safeDeserialize(ResponseInterface $response, $className)
```
Nyholm: Hm, Im not sure about this. There are 3 possible outcomes of an API request. 

1) Success
2) [Common error code](https://documentation.mailgun.com/api-intro.html#errors)
3) Special HTTP status. Like 201 or 202. 

Number 2 should be handled in AbstractApi but 1 and 3 should be handled on each specific resource. 

👎  for this function.

src/Mailgun/Api/AbstractApi.php
```diff
@@ -137,6 +189,60 @@ protected function delete($path, array $parameters = [], array $requestHeaders =
...
+     * Send a request with parameters encoded as multipart-stream form data.
+     *
+     * @param string $type           Request type. (POST, PUT, etc.)
+     * @param string $path           Request path.
+     * @param array  $parameters     POST parameters to be mutipart-stream-encoded.
+     * @param array  $requestHeaders Request headers.
+     *
+     * @return ResponseInterface
+     */
+    protected function doMultipart($type, $path, array $parameters = [], array $requestHeaders = [])
```
Nyholm: Im not a fan of the multipart functions. But I think they are okey for now.

src/Mailgun/Resource/Api/Domain/ComplexDomain.php
```diff
@@ -0,0 +1,93 @@
...
+            $data['domain'],
+            $data['receiving_dns_records'],
+            $data['sending_dns_records']
+        );
+    }
+
+    /**
+     * @param data
+     */
+    public function __construct($domainInfo, $rxRecords, $txRecords)
```
Nyholm: Rather use objects here. 

`__construct(SimpleDomain $domain, DomainDnsRecord $rx, DomainDnsRecord, $tx) `

src/Mailgun/Resource/Api/Domain/ComplexDomain.php
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * @var DomainDnsRecord[]
+     */
+    private $outboundDnsRecords;
+
+    /**
+     * @param array $data
+     *
+     * @return ComplexDomain|array|ResponseInterface
```
Nyholm: This will always return a `ComplexDomain`, right?

Nyholm: Same for all similar occurrences.

src/Mailgun/Resource/Api/Domain/ComplexDomain.php
Nyholm: Same for all similar occurrences.

tests/Integration/DomainApiTest.php
```diff
@@ -0,0 +1,417 @@
...
+     * in the returned list.
+     *
+     * @group domain
+     * @group integration
+     */
+    public function testDomainsList()
+    {
+        $mg = $this->getMailgunClient();
+
+        // @returns DomainListResponse
```
Nyholm: This annotation is not needed. It should be removed.

(It is also in the wrong format. The correct format is `/** @var DomainListReponse $domainList */`)

tests/Integration/DomainApiTest.php
```diff
@@ -0,0 +1,417 @@
...
+    protected function getApiClass()
+    {
+        return 'Mailgun\Api\Domains';
+    }
+
+    /**
+     * Performs `GET /v3/domains` and ensures $this->testDomain exists
+     * in the returned list.
+     *
+     * @group domain
```
Nyholm: Is this group needed?

tests/Api/TestCase.php
```diff
@@ -28,8 +56,20 @@ protected function getApiMock()
...
+                    'post', 'postRaw', 'postMultipart',
+                    'delete', 'deleteMultipart',
+                    'put', 'putMultipart',
+                ]
+            )
             ->setConstructorArgs([$httpClient, $requestClient, $serializer])
             ->getMock();
     }
+
+    protected function getMailgunClient()
```
Nyholm: Maybe have this in a different base class for integration tests?

tests/Integration/DomainApiTest.php
```diff
@@ -0,0 +1,417 @@
...
+     * @group integration
```
Nyholm: I think this could be specified in phpunit.xml.dist. You could say that all tests in a directory are integration tests.

tests/Integration/DomainApiTest.php
```diff
@@ -0,0 +1,417 @@
...
+     * Performs `POST /v3/domains/<domain>/credentials` to attempt to add an invalid
+     * credential pair.
+     *
+     * @expectedException InvalidArgumentException
+     *
+     * @group domain
+     * @group integration
+     * @group negative
+     */
+    public function testCreateCredential_BadPassword_Long()
```
Nyholm: testCreateCredentialBadPasswordLong

.travis.yml
pirogoeth: Travis does not export the secret environment variables when running PR builds. Since this could cause confusion, it's better to output a message than just do nothing or let the integration tests on the PR fail.

Nyholm: > Travis does not export the secret environment variables when running PR builds.

Okey, I did not know. Cool

composer.json
DavidGarciaCat: I thought that PHPUnit was offering "by default" a way to build a code coverage report, even if that is not the case, a code coverage tool - IMHO - should be a local vendor and never thought about it as "the panacea" given some cases may not cover 100% and we really want to do that. given my experience, the only way to validate our code - 100% - is writing tests **before** to add a single line of code. I don't know if is a good idea to add this vendor, even for a `dev` environment... anyway, this is just my opinion...

src/Mailgun/Api/AbstractApi.php
DavidGarciaCat: Most of HTTP Clients are offering a `->isSuccessful()` method and it checks if the status code is `>= 200 && < 300` so maybe - not sure yet - we could use this method (or this `if` statement) to provide the expected behaviour. What is your thought about this?

src/Mailgun/Api/Domain.php
```diff
@@ -0,0 +1,270 @@
...
+        Assert::stringNotEmpty($domain);
+        Assert::stringNotEmpty($login);
+
+        $response = $this->delete(
+            sprintf(
+                '/v3/domains/%s/credentials/%s',
+                $domain,
+                $login
+            ),
+            []
```
DavidGarciaCat: If we're going to send empty arrays, I would like to suggest to set an empty array by default for that parameter. It does not make sense to me write `[]` given it means "we are providing... (drumroll) Nothing!"

src/Mailgun/Resource/Api/Domain/DeliverySettingsUpdateResponse.php
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * @param array $data
+     *
+     * @return SettingsUpdateResponse|array|ResponseInterface
+     */
+    public static function createFromArray(array $data)
+    {
+        $message = array_key_exists('message', $data) ? $data['message'] : null;
+        $noVerify = array_key_exists('skip_verification', $data) ? $data['skip_verification'] : null;
+        $requireTLS = array_key_exists('require_tls', $data) ? $data['require_tls'] : null;
```
DavidGarciaCat: ❓ Why are not setting here a `bool` as by default value, instead of `null`? IMO should be nice if the SDK provides the best/secure option by default and `null` is not giving value from my POV.

src/Mailgun/Resource/Api/Domain/DomainDnsRecord.php
```diff
@@ -0,0 +1,141 @@
...
+    {
+        return $this->priority;
+    }
+
+    /**
+     * @return bool
+     */
+    public function isValid()
+    {
+        return ($this->valid == 'valid') ? true : false;
```
DavidGarciaCat: if we need to check a string and the value should be `valid` always then I suggest this:
`return 'valid' === $this->valid ? true : false`
Strict comparision & Yoda condition - Also, as a PSR Standard, I think we don't need to use `(` and `)` here.

tests/Api/StatsTest.php
```diff
@@ -29,14 +29,15 @@ public function testTotal()
...
-    //}
+    // /**
+    //  * @expectedException \Mailgun\Exception\InvalidArgumentException
+    //  */
+    // public function testTotalInvalidArgument()
+    // {
+    //     $api = $this->getApiMock();
+    //
+    //     $api->total('');
+    // }
```
DavidGarciaCat: As a good practice, commented code should not be committed.
We can check the file history if we need to check something.

tests/Api/StatsTest.php
```diff
@@ -53,12 +54,13 @@ public function testAll()
...
-    //}
+    // /**
+    //  * @expectedException \Mailgun\Exception\InvalidArgumentException
+    //  */
+    // public function testAllInvalidArgument()
+    // {
+    //     $api = $this->getApiMock();
+    //
+    //     $api->all('');
+    // }
```
DavidGarciaCat: Same case here

tests/Api/TestCase.php
```diff
@@ -28,8 +56,20 @@ protected function getApiMock()
             ->getMock();
 
         return $this->getMockBuilder($this->getApiClass())
-            ->setMethods(['get', 'post', 'postRaw', 'delete', 'put'])
+            ->setMethods(
+                [
+                    'get',
+                    'post', 'postRaw', 'postMultipart',
+                    'delete', 'deleteMultipart',
+                    'put', 'putMultipart',
```
DavidGarciaCat: maybe I am wrong, but AFAIK method names should be uppercase, right?

tests/Integration/DomainApiTest.php
DavidGarciaCat: More cases like this below

src/Mailgun/Resource/Api/Domain/DomainDnsRecord.php
Nyholm: I agree. But you could make it even simpler by: 
return 'valid' === $this->valid;

tests/Api/StatsTest.php
Nyholm: Sorry. My bad. I made that mistake.  That is a test that should be used when asset 1.3 is released.

tests/Api/TestCase.php
Nyholm: No. this looks correct.

src/Mailgun/Resource/Api/Domain/DomainDnsRecord.php
DavidGarciaCat: @Nyholm that's true 👍

tests/Api/StatsTest.php
DavidGarciaCat: Do we know the release date?

Nyholm: Any day now. As soon as Bernhard finds some time for open source he will tag a new release.

PR #209, opened 2016-10-31
Different API implementation

pirogoeth:
@Nyholm I believe the other API design we tried imposed a lot of overhead when it comes to implementation and maintenance costs. I honestly think we would be better off if we don't do any "extended" deserialization of the API response. Simply returning a `\stdClass` or `array` should be sufficient, in my opinion. Any thoughts?


pirogoeth:
@Nyholm @DavidGarciaCat any comments on this?

DavidGarciaCat:
@pirogoeth IMHO provide helpers or parameters based on known objects helps the developer, specially when the IDE can access to those properties or methods because there is a PHP file that can manage them.

An example of this is [MangoPay SDK](https://github.com/Mangopay/mangopay2-php-sdk/tree/master/MangoPay). They have some [API classes](https://github.com/Mangopay/mangopay2-php-sdk/blob/master/MangoPay/ApiCards.php) to manage the API calls and then some [DTOs](https://github.com/Mangopay/mangopay2-php-sdk/blob/master/MangoPay/Card.php) to manage easily each object (no matter if we send or get data).

Based on this, I know that I will get a Card object when I can a Card API Endpoint (for example), or that I will need to provide a Card object if I want to submit a new Card.

Something that I really appreciate when I use a SDK is how easily I can work with it. If I need to read the SDK docs to know how it works and then the API docs to find the parameters that I need, then maybe I can build my own SDK to integrate the API, based on my needs.

Nyholm:
I know it has taken a lot of effort building the domain models and that we only covered 2 apis of 8(?). As David says, the benefit of domain objects is that the API is self explanatory and developers do like it. 

But I do also agree with Sean. It will be **a lot of work** setting this up. Im back and forth... 

What is the maintenance cost of 40+ value objects? Aren't they just getters and setters? Those value objects needs to be changes whenever Mailgun decides to change their API. 

-------------

When I look at the stats and domain API and compare those with the once we have left I see that we are more than 40% done... Domain API was the real big one.

DavidGarciaCat:
@Nyholm maybe these objects can have public properties, so it reduces the maintenance cost of each getter/setter. However it means we won't be able to validate each value when we set them...

Nyholm:
Na. Public properties are an anti pattern. My point is that I'm not sure value objects increases maintenance cost. They are just a lot to create.

DavidGarciaCat:
I agree

pirogoeth:
> I know it has taken a lot of effort building the domain models and that we only covered 2 apis of 8(?). As David says, the benefit of domain objects is that the API is self explanatory and developers do like it.

I do agree. The domain/resource objects way of doing this is nice in that everything is wrapped up nicely like little presents...

> But I do also agree with Sean. It will be a lot of work setting this up. Im back and forth...

... But I've got to say that in all honesty, I'd prefer only wrapping the input side of the API and letting the user handle the API output (with the possibility of some helper functions, of course). It makes future maintenance cost remarkably lower, and users aren't necessarily relying on us solely to maintain the response processing side of the library (since they're doing it themselves).

Not only does this keep the implementation simpler, but it also makes it easier for other contributors and users to read through the library itself. There is much less to dig through to try to find issues, especially if we have custom deserializing code (`createFromArray` / `createFromObject` / ...) for each resource, etc etc.

> What is the maintenance cost of 40+ value objects? Aren't they just getters and setters? Those value objects needs to be changes whenever Mailgun decides to change their API.

But also custom data unpacking for each value object.. Almost every endpoint covered in domains API has a separate resource object that is returned.. I think that if the Mailgun API had a more uniform return, doing a domain/resource-style API would be much more feasible.

Right now, the only uniform response I've seen is the `SimpleResponse` I implemented in the domains API, which isn't even all that uniform. Sometimes it contains a `spec` (???? - undocumented on API side...), sometimes it contains `retrySeconds` (rate limiting response..), but it always has a `message` (return string).

Nyholm:
> It makes future maintenance cost remarkably lower

Less code: yes. But what is the maintenance cost on an immutable value object?

> users aren't necessarily relying on us solely to maintain the response processing side of the library 

Yes, The client has to do an update whenever the API does. But we discussed that the API was quite stable now, right?

> Not only does this keep the implementation simpler, but it also makes it easier for other contributors and users to read through the library itself. 

I agree, right now it is quite messy because we have not found a good order yet. But I do think we can structure/refactor/rename so we get a better overview over the classes. 

> Almost every endpoint covered in domains API has a separate resource object that is returned.. I think that if the Mailgun API had a more uniform return, doing a domain/resource-style API would be much more feasible.

Yes, almost every endpoint from the API server has different response format... that is a pain =)

What I have found is that each endpoint uses one or two classes. The endpoints that are using two classes returns a list of something. 

I believe that we should not use inheritance at all for our value objects to keep them clean an simple. We could also name them in a good pattern like `FooEndpointResponse` and `FooEndpointListItem`.

PR #217, opened 2016-11-12
Introduce a request builder.

### LINKED ISSUE #205: Review the HttpClientConfigurator
Are we happy with a class like that? It's purpose is to create a HTTPClient. 

Should we load that class with other HTTP related stuff like a RequestFactory and the MultipartStreamBuilder?

Related to https://github.com/mailgun/mailgun-php/issues/186#issuecomment-255882949

Nyholm:
We inject every dependency (eg RequestFactory and MultipartStreamBuilder) and we do not have to use postMultipart.

This will fix #205

### DIFF
--- a/src/Mailgun/RequestBuilder.php
+++ b/src/Mailgun/RequestBuilder.php
@@ -1,0 +1,92 @@
+<?php
+
+namespace Mailgun;
+
+use Http\Discovery\MessageFactoryDiscovery;
+use Http\Message\MessageFactory;
+use Http\Message\MultipartStream\MultipartStreamBuilder;
+use Http\Message\RequestFactory;
+use Psr\Http\Message\RequestInterface;
+
+/**
+ *
+ *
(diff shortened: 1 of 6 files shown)


src/Mailgun/RequestBuilder.php
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class RequestBuilder
+{
+    /**
+     * @var RequestFactory
+     */
+    private $requestFactory;
```
Nyholm: Missing public setter

src/Mailgun/RequestBuilder.php
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * @var MultipartStreamBuilder
+     */
+    private $multipartStreamBuilder;
```
Nyholm: Missing public setter

Issue #225, opened 2016-11-23
Argument validation on API results


Nyholm:
When the API answers with something unexpected we throw an `InvalidArgumentException` on the user. Which means that the user should retry the API call with different parameters. That is wrong. The user did not do anything wrong. It was either the API or our API client that has made some misstakes. 

In my opinion there is two possible solutions: 
 1. Throw an `ServerAnsweredSomethingUnexpectedException`
 2. Return the result as good as we can.

I would prefer 2 which means we remove all assertions of the server response.

pirogoeth:
I also agree with 2. I think that if we just have to return the "best guess" response, we should log it / throw a warning noting that the user should probably file an issue here because `x`, `y`, or `z` is broken.

Nyholm:
Agreed!

Nyholm:
Closed by #230

PR #230, opened 2016-11-24
Do not validate API responses. Create Response objects that have part…

Nyholm:
…ial data instead

Fix issue #225

### DIFF
--- a/src/Mailgun/Resource/Api/Domain/ConnectionResponse.php
+++ b/src/Mailgun/Resource/Api/Domain/ConnectionResponse.php
@@ -9,7 +9,6 @@
 
 namespace Mailgun\Resource\Api\Domain;
 
-use Mailgun\Assert;
 use Mailgun\Resource\ApiResponse;
 
 /**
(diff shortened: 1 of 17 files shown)


Nyholm:
Can I get any feedback on this?

Nyholm:
This PR removes validation of server responses. We did not test these validations before, that's why the tests are unchanged. 

@pirogoeth, are you okey with merging this?

pirogoeth:
I will merge this, but at this point, I believe we should just drop the resource half of our domain-resource paradigm. We should keep this simple - provide a wrapper for users to use to access the API with clean calls, receive clean, untouched JSON output back from the API. No trying to load it in to objects, no validation. If anything, we can provide some utility helpers that will help with common API tasks (pagination, fetching from a links blocks, fetching stored message from the correct API endpoint, etc).

------

Realistically, if we are not going to validate server responses, then we have no need for multiple response classes. If anything we could have one response class containing the pagination common code as well as a \stdClass member (the JSON-decoded response). This makes consuming data from the API easier for the users and ensures that we don't accidentally "protecting" the users from using their data the way they want to.

Nyholm:
We talked about this. I can't find the issue though. 

I created an example in the readme (#240) to show how you can get this clean array/json output if you want. 

What makes the consuming of the API easier is a matter of taste. Ask any Java developer and he/she will always say that one should use objects. A python developer on the other hand would prefer primary types instead. PHP is somewhere in between. 

If we could serve one APIResponse that would be great. But the API is not build that way.

PR #234, opened 2016-11-24
Clean up gitignore.

Nyholm:
.DS_STORE
.idea 
etc should be in the global gitignore. 

smoketest.php and other test files should go in test/.


Nyholm:
Thank you for the review. 
Feel free to merge PRs like this. Especially since they are small and not really significant. =)

Issue #235, opened 2016-11-24
Improve the integration tests


Nyholm:
I believe we have to rethink the integration tests. Everybody should be able to run them without risking that they will mess up their account. 
Is also seams like they are set up to an account that has some predefined state. 

TODOs: 
 * [ ] Write documentation how to run the tests
 * [ ] Do not test server response messages
 * [ ] Maybe integration tests only should be used for key request, like sending emails. 

When I run the test with my credentials I get the following: 
```text
PHPUnit 4.8.29 by Sebastian Bergmann and contributors.

...EFFF..F.F..F..F

Time: 11.6 seconds, Memory: 6.00MB

There was 1 error:

1) Mailgun\Tests\Integration\DomainApiTest::testDomainCreate
Error: Call to undefined method Mailgun\Resource\Api\ErrorResponse::getDomain()

/Users/tobias/Workspace/PHPStorm/httplug/vendor/mailgun/mailgun-php/tests/Integration/DomainApiTest.php:99

--

There were 7 failures:

1) Mailgun\Tests\Integration\DomainApiTest::testDomainCreateDuplicateValues
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'This domain name is already taken'
+'Missing parameter 'name''

/Users/tobias/Workspace/PHPStorm/httplug/vendor/mailgun/mailgun-php/tests/Integration/DomainApiTest.php:120

2) Mailgun\Tests\Integration\DomainApiTest::testRemoveDomain
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'Domain has been deleted'
+'Domain not found'

/Users/tobias/Workspace/PHPStorm/httplug/vendor/mailgun/mailgun-php/tests/Integration/DomainApiTest.php:133

3) Mailgun\Tests\Integration\DomainApiTest::testCreateCredential
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'Created 1 credentials pair(s)'
+'Parameters are missing'

/Users/tobias/Workspace/PHPStorm/httplug/vendor/mailgun/mailgun-php/tests/Integration/DomainApiTest.php:151

4) Mailgun\Tests\Integration\DomainApiTest::testListCredentials
Failed asserting that false is true.

/Users/tobias/Workspace/PHPStorm/httplug/vendor/mailgun/mailgun-php/tests/Integration/DomainApiTest.php:212

[...]

PR #238, opened 2016-12-06
Add Gitter badge to readme

pirogoeth:

PR #242, opened 2016-12-06
Disable integration tests until they are improved

Nyholm:
Until #235 is resolved.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,8 +33,8 @@
     ],
     "scripts": {
         "test": "vendor/bin/phpunit --testsuite unit && vendor/bin/phpunit --testsuite functional",
-        "test-all": "vendor/bin/phpunit",
+        "test-all": "vendor/bin/phpunit --testsuite all",
         "test-integration": "vendor/bin/phpunit --testsuite integration",
-        "test-coverage": "vendor/bin/phpunit --coverage-text --coverage-clover=build/coverage.xml"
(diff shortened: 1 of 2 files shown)


Nyholm:
Haha agreed. But I want to disable them in to make travis test green. Othewise only admin can merge PRs until #235 is resolved..

Example, I would like to merge this: https://github.com/mailgun/mailgun-php/pull/238

DavidGarciaCat:
Why to disable these tests instead of fix the ones that are failing? Please don't misunderstand me, I know what you mean, but from my POV we can improve the tests later, however ignore a known error is not good. Fortunately Mailgun sends emails, but what happen if we work with money? Are you going to ignore that test? :)

Nyholm:
The test was just recently added. Then we did a refactoring and the test broke. It is not the code that is broken but the tests. 
Since I cannot properly run the locally I have a hard time fixing them right now. (More described in #235) 

If the tests are not fixed or disabled I can no longer push anything to this repo because of the current "protection settings".

DavidGarciaCat:
Up to you - As I said on PR #235 best practices suggest to write test first to prevent unexpected behaviours, and here I described the example (car wheels), so if @pirogoeth is happy with this update, then feel free to merge 😄 - @pirogoeth ?

pirogoeth:
I've got no problems with disabling the integration tests for now. If this was a stable release of the bindings, that's one thing, but this is under active development in a not even nearly stable phase. And, it's not like they're going to be disabled forever, just long enough to fix the tests.

PR #245, opened 2016-12-07
Added Event API with pagination

### LINKED ISSUE #188: Create Events API
Includes request dispatcher and any request//response models necessary.

Nyholm:
This will fix #188 

Do you like the pagination?

```php
$result = $mg->event()->get('domain');
$resultNext = $mg->event()->getPaginationNext($result);
```

Below is an alternative. It does not work as well for someone that does not use the domain objects. And it makes the `EventResponse` a service somehow and not a value object. 
```php
$result = $mg->event()->get('domain');
$resultNext = $result->paginationNext($mg->event());
```

### DIFF
--- a/src/Mailgun/Api/Event.php
+++ b/src/Mailgun/Api/Event.php
@@ -1,0 +1,69 @@
+<?php
+
+namespace Mailgun\Api;
+
+use Mailgun\Assert;
+use Mailgun\Resource\Api\Event\EventResponse;
+
+/**
+ * {@link https://documentation.mailgun.com/api-events.html}
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class Event extends HttpApi
+{
+    use Pagination;
+
+    /**
+     * @param string $domain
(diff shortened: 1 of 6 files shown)


Nyholm:
I can't really move the trait into a folder named trait. Since PSR0 and PSR4 require the namespace to be the same as folder structure. I can't have "trait" in the namespace because it is a reserved word. Using "traits" violates the idea that namespace always should be singular. 

Also. I do not have a separate folder for classes or interfaces... I'm not a big fan of moving the trait to a subfolder just because it is a trait. :s

DavidGarciaCat:
Understood - in fact we cannot define a `Interfaces` folder for example due the Namespace cannot manage it properly (AFAIK) so when I need to deal with this then I use to add the Interface or Trait keyword at the end, for example `MyAwesomeFileTrait`. Anyway, as I said, up to you! 😄

pirogoeth:
Fantastic! Great work on the pagination part :)

Nyholm:
Tank you for merging

PR #247, opened 2016-12-07
Messages api

### LINKED ISSUE #168: Retrieving stored messages

Nyholm:
This will add the missing Message API. The old message builders will still work. 

(Sorry for including the change to `$this->safeDeserialize($response, ...)` in this PR.)

Fixes #168


Nyholm:
Thank you for merging

Issue #190, opened 2016-10-14
Create Routes API


pirogoeth:
Includes request dispatcher and any request//response models necessary.

Nyholm:
Closed by #249

PR #249, opened 2016-12-08
Routes API

DavidGarciaCat:
# Not tested yet

Create Routes API #190

https://documentation.mailgun.com/api-routes.html

https://documentation.mailgun.com/user_manual.html#routes

### DIFF
--- a/src/Mailgun/Api/Routes.php
+++ b/src/Mailgun/Api/Routes.php
@@ -106,7 +106,7 @@ class Routes extends HttpApi
      *
      * @return UpdateResponse
      */
-    public function update($routeId, $expression = null, array $actions = null, $description = null, $priority = null)
+    public function update($routeId, $expression = null, array $actions = [], $description = null, $priority = null)
     {
         Assert::stringNotEmpty($routeId);
 
@@ -116,13 +116,11 @@ class Routes extends HttpApi
             $params['expression'] = trim($expression);
         }
 
-        if (is_array($actions)) {
-            foreach ($actions as $action) {
-                Assert::stringNotEmpty($action);
+        foreach ($actions as $action) {
+            Assert::stringNotEmpty($action);
 
-                $params['action'] = isset($params['action']) ? $params['action'] : [];
-                $params['action'][] = $action;
-            }
+            $params['action'] = isset($params['action']) ? $params['action'] : [];
+            $params['action'][] = $action;
         }
 
         if (is_string($description) && 0 < strlen($description)) {
--- a/src/Mailgun/Api/Routes.php
+++ b/src/Mailgun/Api/Routes.php
@@ -38,6 +38,7 @@ class Routes extends HttpApi
         Assert::integer($limit);
         Assert::integer($skip);
         Assert::greaterThan($limit, 0);
+        Assert::greaterThanEq($skip, 0);
 
         $params = [
             'limit' => $limit,
--- a/src/Mailgun/Api/Routes.php
+++ b/src/Mailgun/Api/Routes.php
@@ -15,6 +15,7 @@ use Mailgun\Resource\Api\Routes\Response\CreateResponse;
 use Mailgun\Resource\Api\Routes\Response\DeleteResponse;
 use Mailgun\Resource\Api\Routes\Response\IndexResponse;
 use Mailgun\Resource\Api\Routes\Response\ShowResponse;
+use Mailgun\Resource\Api\Routes\Response\UpdateResponse;
 
 /**
  * {@link https://documentation.mailgun.com/api-routes.html}.
@@ -102,10 +103,38 @@ class Routes extends HttpApi
      * @param string|null $description An arbitrary string
      * @param int|null    $priority    Integer: smaller number indicates higher priority. Higher priority routes are handled first. Defaults to 0.
      *
-     * @return
+     * @return UpdateResponse
      */
     public function update($routeId, $expression = null, array $actions = null, $description = null, $priority = null)
     {
+        Assert::stringNotEmpty($routeId);
+
+        $params = [];
+
+        if (is_string($expression) && 0 < strlen($expression)) {
+            $params['expression'] = trim($expression);
+        }
+
+        if (is_array($actions)) {
+            foreach ($actions as $action) {
+                Assert::stringNotEmpty($action);
+
+                $params['action'] = isset($params['action']) ? $params['action'] : [];
+                $params['action'][] = $action;
+            }
+        }
+
+        if (is_string($description) && 0 < strlen($description)) {
+            $params['description'] = trim($description);
+        }
+
+        if (is_int($priority)) {
+            $params['priority'] = $priority;
+        }
+
+        $response = $this->httpPut(sprintf('/v3/routes/%s', $routeId), $params);
+
+        return $this->safeDeserialize($response, UpdateResponse::class);
     }
 
     /**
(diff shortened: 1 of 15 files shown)


src/Mailgun/Api/Routes.php
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * {@link https://documentation.mailgun.com/api-routes.html}.
+ *
+ * @author David Garcia <me@davidgarcia.cat>
+ */
+class Routes extends HttpApi
+{
+    /**
+     * Fetches the list of Routes. Note that Routes are defined globally,
```
Nyholm: Good. This should be done on other apis as well

src/Mailgun/Api/Routes.php
```diff
@@ -0,0 +1,157 @@
...
+
+        return $this->safeDeserialize($response, IndexResponse::class);
+    }
+
+    /**
+     * Returns a single Route object based on its ID.
+     *
+     * @param string $routeId Route ID returned by the Routes::index() method
+     *
+     * @return RouteDto
```
Nyholm: Should be ShowResponse, right?

src/Mailgun/Api/Routes.php
```diff
@@ -0,0 +1,157 @@
...
+     *
+     * @param string      $routeId     Route ID returned by the Routes::index() method
+     * @param string|null $expression  A filter expression like "match_recipient('.*@gmail.com')"
+     * @param array|null  $actions     Route action. This action is executed when the expression evaluates to True. Example: "forward('alice@example.com')"
+     * @param string|null $description An arbitrary string
+     * @param int|null    $priority    Integer: smaller number indicates higher priority. Higher priority routes are handled first. Defaults to 0.
+     *
+     * @return UpdateResponse
+     */
+    public function update($routeId, $expression = null, array $actions = null, $description = null, $priority = null)
```
Nyholm: Array $axtions = []

src/Mailgun/Api/Routes.php
```diff
@@ -0,0 +1,157 @@
...
+    {
+        Assert::stringNotEmpty($routeId);
```
Nyholm: You can create assertions like Assert::stringOrEmpty

src/Mailgun/Api/Routes.php
```diff
@@ -0,0 +1,157 @@
...
+
+        $params = [];
+
+        if (is_string($expression) && 0 < strlen($expression)) {
```
Nyholm: Use if(empty()) instead

src/Mailgun/Api/Routes.php
```diff
@@ -0,0 +1,157 @@
...
+            $params['expression'] = trim($expression);
+        }
+
+        if (is_array($actions)) {
```
Nyholm: This is not needed if we update the default value

src/Mailgun/Api/Routes.php
```diff
@@ -0,0 +1,157 @@
...
+                $params['action'] = isset($params['action']) ? $params['action'] : [];
+                $params['action'][] = $action;
+            }
+        }
+
+        if (is_string($description) && 0 < strlen($description)) {
+            $params['description'] = trim($description);
+        }
+
+        if (is_int($priority)) {
```
Nyholm: Use empty

src/Mailgun/Api/Routes.php
```diff
@@ -0,0 +1,157 @@
...
+        if (is_array($actions)) {
+            foreach ($actions as $action) {
+                Assert::stringNotEmpty($action);
+
+                $params['action'] = isset($params['action']) ? $params['action'] : [];
+                $params['action'][] = $action;
+            }
+        }
+
+        if (is_string($description) && 0 < strlen($description)) {
```
Nyholm: Use empty

src/Mailgun/Resource/Api/Routes/Dto/RouteDto.php
```diff
@@ -0,0 +1,133 @@
...
+     * @param string $createdAt
+     */
+    private function __construct($id, $priority, $expression, $actions, $description, $createdAt)
+    {
+        $this->id = $id;
+        $this->priority = $priority;
+        $this->filter = $expression;
+        $this->actions = ActionDto::createMultiple($actions);
+        $this->description = $description;
+        $this->createdAt = !is_null($createdAt) ? new \DateTime($createdAt) : $createdAt;
```
Nyholm: Update parameters to make sure the last arg always is a DateTime (or null) and you do not need this

src/Mailgun/Resource/Api/Routes/Dto/RouteDto.php
```diff
@@ -0,0 +1,133 @@
...
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+namespace Mailgun\Resource\Api\Routes\Dto;
+
+/**
+ * @author David Garcia <me@davidgarcia.cat>
+ */
+final class RouteDto
```
Nyholm: What does Dto stand for?

src/Mailgun/Api/Routes.php
DavidGarciaCat: You're right, my mistake when I updated the code - I'll add an update here

src/Mailgun/Api/Routes.php
DavidGarciaCat: Why to use `stringOrEmpty` when I need a string and should not be empty due the Route ID is used on the API Call as part of the URL? It sounds to me that I need to be sure its a string and is not empty.

src/Mailgun/Api/Routes.php
DavidGarciaCat: I presume you mean `if (!empty($expression)) {` right? If I check just if the var is empty or not then I can pass `$expression = ['asdfg', 'qwerty']` and we won't get an error, when it's obvious that is wrong. This is why I am checking if we have an empty.

If you mean `if (is_string($expression) && !empty($expression))` then 👍

src/Mailgun/Api/Routes.php
DavidGarciaCat: Already removed - Coming soon when push new updates

src/Mailgun/Api/Routes.php
DavidGarciaCat: Same case described above - Can you confirm please?

src/Mailgun/Api/Routes.php
DavidGarciaCat: Same case. Priority must be an integer, based on the API docs, so `$priority = 'qwerty';` is not empty (we will get `true` BUT the value is not right due "qwerty" is a string instead of a number. Are you 100% sure about this?

src/Mailgun/Api/Routes.php
Nyholm: I meant `if(!empty($expression)) {`. It makes sense with my previous comment https://github.com/mailgun/mailgun-php/pull/249#discussion_r91692461

src/Mailgun/Api/Routes.php
Nyholm: https://github.com/mailgun/mailgun-php/pull/249#discussion_r91692674

`if(!empty($description)) {`

src/Mailgun/Api/Routes.php
Nyholm: Yes. But you must use `Assert::nullOrInteger()` before.

PR #250, opened 2016-12-09
Minor amends reported by Scrutinizer CI

DavidGarciaCat:
A set of minor updates reported by Scrutinizer CI.

### DIFF
--- a/src/Mailgun/Resource/Api/Message/ShowResponse.php
+++ b/src/Mailgun/Resource/Api/Message/ShowResponse.php
@@ -339,14 +339,6 @@ class ShowResponse implements ApiResponse
     }
 
     /**
-     * @param array $attachments
-     */
-    private function setAttachments(array $attachments)
-    {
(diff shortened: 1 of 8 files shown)


src/Mailgun/Resource/Api/Message/ShowResponse.php
```diff
@@ -371,14 +363,6 @@ public function getContentIdMap()
     }
 
     /**
-     * @param string $contentIdMap
-     */
-    private function setContentIdMap($contentIdMap)
```
Nyholm: Not sure this should be removed, it should be used instead... =)

Same with setAttachment.

Nyholm: I agree that we should not have dead code.. 
It was a bug that those functions are not used. Removing those functions does not resolve the bug. I would much rather see those functions to still be there and we fix the bug.

Issue #254, opened 2016-12-20
Adding custom header - add a non-array value


jameskraus:
Right now adding a custom header adds the value as an array:

```
    public function addCustomHeader($headerName, $headerData)
    {
        if (!preg_match('/^h:/i', $headerName)) {
            $headerName = 'h:'.$headerName;
        }
        $this->message[$headerName] = [$headerData];

        return $this->message[$headerName];
    }
```

Why isn't it just:

`$this->message[$headerName] = $headerData;`

Because right now I get weird headers like this:

`X-Auto-Response-Suppress[0]: All`

When I call it like this:

`$thing->addCustomHeader('X-Auto-Response-Suppress', 'All');`

ivanjaros:
Just encountered this issue myself. Yeah, that should be string since docs expect string, not array: https://documentation.mailgun.com/api-sending.html#sending

jameskraus:
PR was merged.

PR #255, opened 2016-12-20
Bug fix - Allow adding custom headers with non-array value

jameskraus:
Fix for issue https://github.com/mailgun/mailgun-php/issues/254

### DIFF
--- a/src/Mailgun/Messages/MessageBuilder.php
+++ b/src/Mailgun/Messages/MessageBuilder.php
@@ -248,7 +248,16 @@ class MessageBuilder
         if (!preg_match('/^h:/i', $headerName)) {
             $headerName = 'h:'.$headerName;
         }
-        $this->message[$headerName] = [$headerData];
+
+        if (array_key_exists($headerName, $this->message)) {
+            if (is_array($this->message[$headerName])) {
+                $this->message[$headerName][] = $headerData;
+            } else {
+                $this->message[$headerName] = [$this->message[$headerName], $headerData];
+            }
+        } else {
+            $this->message[$headerName] = $headerData;
+        }
 
         return $this->message[$headerName];
     }
--- a/tests/Functional/MessageBuilderHeaderTest.php
+++ b/tests/Functional/MessageBuilderHeaderTest.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Mailgun\Tests\Functional;
+
+/**
+ * @author James Kraus <jkraus@imagineteam.com>
+ */
+class MessageBuilderHeaderTest extends \PHPUnit_Framework_TestCase
+{
+    public function testSimpleExample()
+    {
+        $messageValidator = function ($headers) {
+            $this->assertContains(['name' => 'h:My-Singular-Header', 'contents' => '123'], $headers);
+            $this->assertContains(['name' => 'h:My-Plural-Header[0]', 'contents' => '123'], $headers);
+            $this->assertContains(['name' => 'h:My-Plural-Header[1]', 'contents' => '456'], $headers);
+        };
+
+        // Create the mocked mailgun client.
+        $mailgun = MockedMailgun::create($this, 'POST', 'domain/messages', [], $messageValidator);
+
+        $builder = $mailgun->MessageBuilder();
+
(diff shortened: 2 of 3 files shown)


Nyholm:
Thank you for this PR. I like that you added tests. But I believe we need more functional tests for this. I honestly do not know if this really fixes the problem or just introducing more bugs. 😕  Even though it looks right. 

I would like to have a two or more tests when you are using the MessageBuilder similar to this: 

```php
class MessageBuilderHeaderTest extends \PHPUnit_Framework_TestCase
{
    public function testSimpleExample()
    {
        // Create a Closure that validates the $files parameter to RestClient::send()
        $headersValidator = function($headers) {
            $this->assertContains(['name'=>'My-Header',   'contents'=>'123'], $headers);
            $this->assertContains(['name'=>'My-Header2',  'contents'=>['123', '456']], $headers);
        };

        // Create the mocked mailgun client. 
        $mailgun = MockedMailgun::create($this, 'POST', 'domain/messages', [], [], $headersValidator);

        // TODO Build message with message builder
       // ...
        $mailgun->sendMessage('domain', $messageBuilder->getMessage());
    }
}
```

Have a look at https://github.com/mailgun/mailgun-php/blob/master/tests/Functional/SendMessageTest.php and https://github.com/mailgun/mailgun-php/blob/master/tests/Functional/NoSamePostNameTest.php for inspiration.

jameskraus:
I gave it a whack, but I could only get a functional test to work by passing the headersValidator into the filesValidator parameter of the MockedMailgun::create call.


e.g.
```
class MessageBuilderHeaderTest extends \PHPUnit_Framework_TestCase
{
    public function testSimpleExample()
    {
        $headersValidator = function ($headers) {
            $this->assertContains(['name' => 'h:My-Singular-Header', 'contents' => '123'], $headers);
            $this->assertContains(['name' => 'h:My-Plural-Header[0]', 'contents' => '123'], $headers);
            $this->assertContains(['name' => 'h:My-Plural-Header[1]', 'contents' => '456'], $headers);
        };

        // Create the mocked mailgun client.
        $mailgun = MockedMailgun::create($this, 'POST', 'domain/messages', [], $headersValidator);

        $builder = $mailgun->MessageBuilder();

        $builder->addCustomHeader('My-Singular-Header', '123');
        $builder->addCustomHeader('My-Plural-Header', '123');
        $builder->addCustomHeader('My-Plural-Header', '456');

        $builder->setFromAddress('from@example.com');
        $builder->addToRecipient('to@example.com');
        $builder->setSubject('Foo');
        $builder->setTextBody('Bar');

        $mailgun->sendMessage('domain', $builder->getMessage(), []);
    }
}
```

Nyholm:
Awesome. I tried to run the tests without the patch The functional tests shows that this is indeed a bug.


@DavidGarciaCat: About where to put the test, Im not sure what the best practices are. Either way is fine to me.

src/Mailgun/Messages/MessageBuilder.php
```diff
@@ -248,7 +248,16 @@ public function addCustomHeader($headerName, $headerData)
...
+
+        if (array_key_exists($headerName, $this->message)) {
+            if (is_array($this->message[$headerName])) {
+                $this->message[$headerName][] = $headerData;
+            } else {
+                $this->message[$headerName] = [$this->message[$headerName], $headerData];
+            }
+        } else {
+            $this->message[$headerName] = $headerData;
+        }
```
Nyholm: We could simplify this code bock by always using an array for the headers, even though it is just a single value.  

Wouldn't it also work if this code block was replaced with: 
`$this->message[$headerName][] = $headerData;`

DavidGarciaCat: It should work, but then re-write the PHPUnit asserts will be mandatory, to ensure we can count the number of custom headers (to validate we have the number that we provided), validate the headers keys (adding the `h:` prefix on each one) and the stored values for each header.

Just in case its needed, here's a list of all PHPUnit assertions:
https://phpunit.de/manual/current/en/appendixes.assertions.html

Nyholm:
The current implementation preserves the functionality that we currently have. IE allow singular headers to be singluar and not treated as arrays. 

Thank you @jameskraus  for this PR.

PR #271, opened 2017-01-28
Add getPagination* concrete functions on Pagination trait

pirogoeth:


src/Mailgun/Api/Pagination.php
```diff
@@ -22,6 +23,51 @@
...
+     * @return string
+     */
+    abstract protected function getPaginationBase();
+
+    /**
+     * @param PaginationResponse $response
+     *
+     * @return mixed|null
+     */
+    public function getPaginationNext(PaginationResponse $response)
```
Nyholm: PaginationResponse is a Trait. You cannot use Traits like this. 
https://3v4l.org/K4mdT

src/Mailgun/Api/Pagination.php
```diff
@@ -22,6 +23,51 @@
...
+    {
+        return $this->getPaginationUrl($response->getNextUrl(), $this->getPaginationBase());
```
Nyholm: Why not use `get_class($response);`?

pirogoeth:
@Nyholm addressed your comments, sorry for the delay :)

src/Mailgun/Api/Pagination.php
```diff
@@ -10,6 +10,7 @@
 namespace Mailgun\Api;
 
 use Mailgun\Assert;
+use Mailgun\Resource\Api\mixed;
```
DavidGarciaCat: 👎 this is not a valid resource that we can import. If we can use more than one based on "criteria" then I suggest to add `use Mailgun\Resource\Api;` and other annotations and/or type hint can use `Api\...`

src/Mailgun/Api/Pagination.php
```diff
@@ -22,12 +23,52 @@
     abstract protected function safeDeserialize(ResponseInterface $response, $className);
 
     /**
+     * @param mixed $response
+     *
+     * @return mixed|null
```
DavidGarciaCat: `mixed` type hint (s)

src/Mailgun/Api/Pagination.php
DavidGarciaCat: Plan B is to set a generic Interface that should be implemented by all possible objects that we are going to expect.

pirogoeth: Whoops, didn't mean to do that! Did a find-and-replace on PaginationResponse to get rid of the type hints and set the annotations to mixed, but I guess I missed one :) nice catch!

src/Mailgun/Api/Pagination.php
pirogoeth: Uhh, what do you mean?

src/Mailgun/Api/Pagination.php
```diff
@@ -22,12 +22,52 @@
...
+     */
+    public function getPaginationNext($response)
```
Nyholm: We should still try to type hint this. 
What do you think of adding an interface that cover the `get*Url()` functions from the trait?

src/Mailgun/Api/Pagination.php
```diff
@@ -22,12 +22,52 @@
...
+    {
+        return $this->getPaginationUrl($response->getNextUrl(), get_class($response));
+    }
+
+    /**
+     * @param mixed $response
+     *
+     * @return mixed|null
+     */
+    public function getPaginationPrevious($response)
```
Nyholm: Just an idea: What do you think of renaming the functions to `paginationPrevious`?

pirogoeth: I certainly see no problem with that.

src/Mailgun/Api/Pagination.php
pirogoeth: Yeah, that could work quite well. I'll take care of that on Monday.

src/Mailgun/Api/Pagination.php
DavidGarciaCat: Example:
When we call the API endpoint then the SDK should return a Response, right?

What I mean - ideally - we should type hint the expected content, and `mixed` has no sense to me when I am using a SDK that is managing an API resource. From my POV a SDK should return always known resources, no matter if it is an object, an array, a string...

src/Mailgun/Api/Pagination.php
DavidGarciaCat: I would say `previous` or `previousPage` due we are already using a Pagination object does not make sense to "persist" using `pagination...`

pirogoeth: Or even this ^^.

I like just being able to do `$resp->previous()` and so on, as it's very clean and not redundant.

DavidGarciaCat: Agree

Nyholm: Are we sure that no API will contain ever a call called "next", "previus", "first" and "last"?

DavidGarciaCat: Well, that is why I suggested `previousPage()` for example, but in the other hand, we are working with a Pagination object so make sense to use a method name like `previous()`. Feel free to choose, as far it works as expected.

pirogoeth: I am renaming to `(previous|next|first|last)Page`.

Issue #276, opened 2017-02-13
Getting attachment doesn't work


s22su:
Currently Mailgun API provides attachments as an array with full attachment URL.

Example:

`
[
  {
    "url": "https://se.api.mailgun.net/v3/domains/XXX/messages/XXX/attachments/0",
    "content-type": "image/jpeg",
    "name": "IMG_123.jpg",
    "size": 1545628
  }
]
`

It is not possible to get attachment using current RestClient::get method as it adds API URL (api.mailgun.net) and API version to the URL so it becomes:
https://api.mailgun.net/v2/https://se.api.mailgun.net/v3/domains/XXX/messages/XXX/attachments/0

Possible fixes:

1. change RestClient::get method
2. create a new method for getting an attachment

Fix: #277

s22su:
Closed. See #277 for information.

Nyholm:
Closed by #277 

Thank you @s22su

PR #277, opened 2017-02-13
#276 additional method to get an attachment

s22su:


s22su:
@DavidGarciaCat yes. Will deal with that soon..

Nyholm:
Thank you for this PR. 
The library is in the last stages to be rewritten. The RestClient is one of the classes that will be deprecated within the coming weeks. Just an FYI. The [new API's](https://github.com/mailgun/mailgun-php/milestone/1) should also support this feature. (That is nothing you have to do in this PR)

s22su:
@Nyholm let's close this and #276 then and wait until better API is ready.

Nyholm:
No, we can still merge this in. My personal idea is that version 3.0 is at least a year away. Just because a class is deprecated doesn't mean it will be removed anytime soon. 

I think it would be a good idea to support this feature for people using the "old way".

s22su:
@Nyholm @DavidGarciaCat added this method also as a public method to Mailgun class and created test cases

tests/MailgunTest.php
```diff
@@ -49,4 +49,22 @@ public function testVerifyWebhookEmptyRequest()
...
+
+        $this->assertInstanceOf('stdClass', $response);
+        $this->assertEquals($response->http_response_code, 200);
+    }
+
+    public function testGetAttachmentFail()
+    {
+        $this->setExpectedException('\\Mailgun\\Connection\\Exceptions\\GenericHTTPError');
+        $attachmentUrl = 'https://api.mailgun.net/non.existing.uri/1/2/3';
+        $client = new Mailgun('key-3ax6xnjp29jd6fds4gc373sgvjxteol0');
```
DavidGarciaCat: Is this your key? 😕 TBH I am not sure if is wise to provide the key as plain text here...

s22su: @DavidGarciaCat of course it is not my key. If you look a few lines up in the file then you can see the same key (line 26: https://github.com/mailgun/mailgun-php/blob/master/tests/MailgunTest.php#L26).

Also mailgun documentation uses this key in examples.

DavidGarciaCat: 👍 Thanks for clarify - I did the review out of home and had no time to double-check it.

s22su:
Any chance to have it merged?

Nyholm:
Yes, I'll merge this as a temporary fix. This feature will be replaced in the next major version.

PR #278, opened 2017-02-15
Add quotes around the boundary.

Nyholm:
This will support bounderies with non whitespace


Nyholm:
Thank you for the review

PR #279, opened 2017-02-15
Fix disordered POST parameters

z38:
When sending `multipart/form-data` requests, Mailgun seems to not pick up the order of the parameters correctly. While this is not a problem in most cases, when combining a `stop()` action with other actions to build a route it does matter.

This problem could be solved on Mailgun's side, however I think it makes more sense to send standard-compliant (?) requests in the first place. Please feel free to close this PR, if you think otherwise. I opened an issue at Mailgun to see what they can/will do (but haven't heard back from them yet).


src/Mailgun/Connection/RestClient.php
```diff
@@ -147,6 +147,9 @@ public function post($endpointUrl, array $postData = [], $files = [])
                     $postDataMultipart[] = [
                         'name' => sprintf('%s[%d]', $key, $index++),
                         'contents' => $subValue,
+                        'headers' => [
+                            'Content-Disposition' => sprintf('form-data; name="%s"', $key),
```
Nyholm: Just using `$key` here will not be good. If one adds multiple resources named "foobar" it will be overwritten because the server will only see one resource named "foobar". That is why we need the array syntax.

z38: We used this fix to use the `/v2/routes` endpoint (and deployed it succesfully in production). Are there any endpoints which do not support multiple resources with the same name?

Nyholm: Im not sure. I do not work with the Mailgun api team. (I do not work at Mailgun at all..) I just know it is good practice to use the array syntax instead. **All** HTTP servers and server languages will support the array syntax.

z38: Well, the Mailgun server kind of understands it (it receives all arguments but they are disordered). I think it's best if we wait for Mailgun (either by responding to my support ticket or here) to clear things up. All the other clients/libraries I checked (curl, Go, ...) do not use the array syntax.

Nyholm: Im afraid that the argument "others do not follow standards or best practices" is not a very good one.

z38: As far as I understood, the automatic decoding of square brackets in form names is a specialty of PHP. The applicable [RFC 7578](https://tools.ietf.org/html/rfc7578) does not impose such constraints on names nor does it advise to use brackets to encode arrays.

z38:
I just heard back from Mailgun: They were able to reproduce the issue and suggest that we fix this within this SDK.

With the new version of `php-http/multipart-stream-builder`, the fix is a lot cleaner than before. Please let me know if anything needs to be fixed.

tests/Functional/FileFromMemoryTest.php
```diff
@@ -33,9 +33,9 @@ public function testAddFileFromMemory()
                 ];
             }, $attachments);
 
-            $this->assertContains(['name' => 'attachment[0]', 'contents' => 'File content', 'filename' => 'file1.txt'], $attachments);
-            $this->assertContains(['name' => 'attachment[1]', 'contents' => 'File content 2', 'filename' => 'file2.txt'], $attachments);
-            $this->assertContains(['name' => 'attachment[2]', 'contents' => 'Contents of a text file', 'filename' => 'text_file.txt'], $attachments);
+            $this->assertContains(['name' => 'attachment', 'contents' => 'File content', 'filename' => 'file1.txt'], $attachments);
+            $this->assertContains(['name' => 'attachment', 'contents' => 'File content 2', 'filename' => 'file2.txt'], $attachments);
+            $this->assertContains(['name' => 'attachment', 'contents' => 'Contents of a text file', 'filename' => 'text_file.txt'], $attachments);
```
Nyholm: @pirogoeth Can you verify with the Mailgun API team that this is okey? I thought that we needed the array sytax. 

If this is the preferred format, please merge this PR.

z38:
@DavidGarciaCat The merge conflict should be resolved now.

Nyholm:
If this is merge, please also merge #294 which also removes the array syntax.

tests/Functional/FileFromMemoryTest.php
pirogoeth: @Nyholm Yeah, I will look in to it.

Nyholm:
@pirogoeth Do you have any updates on this?

pirogoeth:
@Nyholm Yep, both the array syntax `attachments[]` and non-array syntax `attachments` should work equivalently. Merging.

PR #280, opened 2017-02-15
RFC: Fix inconsistent names of route properties

z38:
As most of the documentation and all methods use the term `expression` for a route's filter expression, the use of the term "filter" seems a bit counter-intuitive to me.

### DIFF
--- a/src/Mailgun/Resource/Api/Routes/Route.php
+++ b/src/Mailgun/Resource/Api/Routes/Route.php
@@ -27,7 +27,7 @@ final class Route
     /**
      * @var string
      */
-    private $filter;
+    private $expression;
 
     /**
      * @var Action[]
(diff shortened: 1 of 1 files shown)


Nyholm:
Thank you @z38 for this PR. I agree with you from the perspective of the library. We should use a common name for the same thing. (expression/filter)

But as @DavidGarciaCat says: The API server are using the name "filter" and the client should use the same (but weird) terminology.

z38:
Thank you for taking a look at this. I definitely agree with you that "filter" is the more expressive term. Would you be interested in a PR for [renaming](https://github.com/z38/mailgun-php/commit/0af649affdc83697429c26da22c5e1966a654263) the remaining `$expression` arguments to `$filter`?

Nyholm:
The API client is using the names from the API server. (See documentation https://documentation.mailgun.com/user_manual.html). The API server has inconsistent naming, that is why this library has inconsistent naming. 

We are consistent with the server naming.

PR #282, opened 2017-02-15
Avoid fatal error when deserializing unauthorized requests

z38:
In some cases, the API returns a response without any content (e.g. when supplying an invalid key).

Example:
```php
$mailgun = new Mailgun\Mailgun('invalidkey');
$mailgun->routes()->index(); // throws error/exception
```

In such cases, a fatal error was thrown because `null` was passed to `ErrorResponse::create()` (PHP 5.6) or an exception was thrown when the empty body was decoded ([PHP >7.0](https://3v4l.org/IalIO)).

It might be necessary to include a similar fix for `ArrayDeserializer` as well.

### DIFF
--- a/src/Mailgun/Api/HttpApi.php
+++ b/src/Mailgun/Api/HttpApi.php
@@ -12,6 +12,7 @@ namespace Mailgun\Api;
 use Http\Client\Exception as HttplugException;
 use Http\Client\HttpClient;
 use Mailgun\Deserializer\ResponseDeserializer;
+use Mailgun\Exception\HttpClientException;
 use Mailgun\Exception\HttpServerException;
 use Mailgun\RequestBuilder;
 use Mailgun\Resource\Api\ErrorResponse;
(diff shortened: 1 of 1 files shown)


src/Mailgun/Deserializer/ModelDeserializer.php
```diff
@@ -33,10 +33,13 @@ public function deserialize(ResponseInterface $response, $class)
         if (strpos($contentType, 'application/json') !== 0 && strpos($contentType, 'application/octet-stream') !== 0) {
             throw new DeserializeException('The ModelDeserializer cannot deserialize response with Content-Type: '.$contentType);
         }
-
-        $data = json_decode($body, true);
-        if (JSON_ERROR_NONE !== json_last_error()) {
-            throw new DeserializeException(sprintf('Error (%d) when trying to json_decode response', json_last_error()));
+        if ($body !== '') {
```
DavidGarciaCat: I prefer to use an improved validation like `is_string($body) && !empty($body)`

Nyholm:
What response will return no content?

z38:
For example when you supply an invalid key, an empty `401 Unauthorized` will be sent back.

src/Mailgun/Deserializer/ModelDeserializer.php
z38: Fixed.

Nyholm:
We should catch that in "safeDeserialze".

z38:
Should `safeDeserialize` just throw an exception from `Mailgun\Exception\HttpClientException` or should we build and return an `ErrorResponse`?

Nyholm:
I think it should throw a InvalidApiKeyException.

DavidGarciaCat:
Agree with @Nyholm here, an Exception makes more sense to me

z38:
Thank you for reviewing. With the updated patch, unauthorized requests throw an exception now.

PR #286, opened 2017-02-18
Tag api

### LINKED ISSUE #257: Add tags API
https://documentation.mailgun.com/api-tags.html#tags

Nyholm:
This will fix #257

### DIFF
--- a/src/Mailgun/Api/Tag.php
+++ b/src/Mailgun/Api/Tag.php
@@ -22,7 +22,7 @@ use Psr\Http\Message\ResponseInterface;
  *
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
  */
-class Tag extends HttpApi
+final class Tag extends HttpApi
 {
     /**
      * Returns a list of tags.
@@ -102,9 +102,7 @@ class Tag extends HttpApi
         Assert::stringNotEmpty($tag);
 
         $response = $this->httpGet(sprintf('/v3/%s/tags/%s/stats', $domain, $tag), $params);
-
-        // TODO catch 400.
-
+        
         return $this->safeDeserialize($response, StatisticsResponse::class);
     }
 
--- a/src/Mailgun/Api/Tag.php
+++ b/src/Mailgun/Api/Tag.php
@@ -1,0 +1,121 @@
+<?php
+
+/*
+ * Copyright (C) 2013-2016 Mailgun
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+namespace Mailgun\Api;
+
+use Mailgun\Assert;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * {@link https://documentation.mailgun.com/api-tags.html#tags}.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class Tag extends HttpApi
+{
+    /**
+     * Returns a list of tags.
+     *
+     * @param string $domain
+     * @param int    $limit
+     *
+     * @return IndexResponse|ResponseInterface
+     */
+    public function index($domain, $limit = 100)
+    {
+        Assert::stringNotEmpty($domain);
+        Assert::integer($limit);
+
+        $params = [
+            'limit' => $limit,
+        ];
+
+        $response = $this->httpGet(sprintf('/v3/%s/tags', $domain), $params);
+
+        return $this->safeDeserialize($response, IndexResponse::class);
+    }
+
+    /**
+     * Returns a single tag.
+     *
+     * @param string $domain Name of the domain
+     * @param string $tag
+     *
+     * @return ShowResponse|ResponseInterface
+     */
+    public function show($domain, $tag)
+    {
+        Assert::stringNotEmpty($domain);
+        Assert::stringNotEmpty($tag);
+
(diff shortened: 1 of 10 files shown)


src/Mailgun/Api/Tag.php
```diff
@@ -0,0 +1,126 @@
...
+     * @param string $tag
+     * @param string $description
+     *
+     * @return UpdateResponse|ResponseInterface
+     */
+    public function update($domain, $tag, $description)
+    {
+        Assert::stringNotEmpty($domain);
+        Assert::stringNotEmpty($tag);
+        Assert::string($description);
```
DavidGarciaCat: If we are going to accept an empty string, then I would say to set the by-default value:
`public function update($domain, $tag, $description = '')`

src/Mailgun/Api/Tag.php
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * Returns statistics for a single tag.
+     *
+     * @param string $domain Name of the domain
+     * @param string $tag
+     *
+     * @return StatisticsResponse|ResponseInterface
+     */
+    public function stats($domain, $tag, array $params)
```
DavidGarciaCat: Missed `@param array $params` annotation here

DavidGarciaCat: Although the type hinting should already prevent this, IMHO could be a wise idea to add another `Assert` to validate that `$params` is an array, specially when some people doesn't display errors/exceptions (this is usually on Prod environments). Anyway, I believe this is a minor update.

src/Mailgun/Resource/Api/Tag/StatisticsResponse.php
```diff
@@ -0,0 +1,128 @@
...
+     */
+    private $tag;
+    /**
+     * @var string
+     */
+    private $description;
+    /**
+     * @var string
+     */
+    private $resolution;
```
DavidGarciaCat: Do you mind to add an empty line between each property and the next annotation please? Although this is a minor update and has no effect, I believe it helps to improve the readability.

src/Mailgun/Resource/Api/Tag/DeleteResponse.php
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * @param string $message
+     */
+    private function __construct($message)
+    {
+        $this->message = $message;
+    }
+
+    public static function create(array $data)
```
DavidGarciaCat: Missed annotation here. Can you add it please? Thanks,

src/Mailgun/Resource/Api/Tag/IndexResponse.php
```diff
@@ -0,0 +1,54 @@
...
+     * @param Tag[] $items
+     * @param array $paging
+     */
+    public function __construct(array $items, array $paging)
+    {
+        $this->items = $items;
+        $this->paging = $paging;
+    }
+
+    public static function create(array $data)
```
DavidGarciaCat: Same case here - missed annotation. Thanks,

src/Mailgun/Resource/Api/Tag/UpdateResponse.php
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * @param string $message
+     */
+    private function __construct($message)
+    {
+        $this->message = $message;
+    }
+
+    public static function create(array $data)
```
DavidGarciaCat: Another missed annotation 😄

src/Mailgun/Api/Tag.php
Nyholm: I was considering that. But running `update($domain, $tag)` would mean that you erase the description. That is not a obvious consequence. However, doing `update($domain, $tag, '')` would be more obvious.

src/Mailgun/Resource/Api/Tag/StatisticsResponse.php
Nyholm: Of course. I thought that was fixed by StyleCI.. :/

src/Mailgun/Api/Tag.php
DavidGarciaCat: OK - Suggestion:
1. Set-up an empty string: `public function update($domain, $tag, $description = '')`
2. Check if the content is a string `Assert::string($description);` no matter if there is content or not
3. Replace the way we're building the `$params` argument from:
```
$params = [
    'description' => $description,
];
```
to
```
$params = empty(trim($description)) ? [] : ['description' => $description];
```
Makes sense now?

Nyholm: no, because we want to allow users to clear the description. Changing the description is the only reason to use the `update` method.

DavidGarciaCat: OK I see your point - Then is OK

src/Mailgun/Resource/Api/Tag/IndexResponse.php
```diff
@@ -0,0 +1,59 @@
...
+
+namespace Mailgun\Resource\Api\Tag;
+
+use Mailgun\Resource\Api\PaginationResponse;
+use Mailgun\Resource\ApiResponse;
+
+/**
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+final class IndexResponse implements ApiResponse
```
pirogoeth: Can you add the `Mailgun\Resource\Api\PagingProvider` interface here?

PR #291, opened 2017-02-20
Show that we allow MultipartStreamBuilder 0.2

Nyholm:


Nyholm:
We do not allow 0.x. We currently only allow 0.1.x. Semver is behaving differently before 1.0.0.

See https://developer.happyr.com/always-use-caret-instead-of-tilde

Issue #289, opened 2017-02-18
Move models to new namespace


Nyholm:
Move them from `Resources\Api` to `Model`

MichalHanajik:
Sent PR #293

Issue #288, opened 2017-02-18
Make sure all models are final


Nyholm:

MichalHanajik:
Any tips which models are pretty sure extendable?
Marking all of them makes all [tests pass](https://travis-ci.org/MichalHanajik/mailgun-php/builds/203978868) 😄

Nyholm:
Hm... Then it is okey =)

MichalHanajik:
I'll send it as PR and then we will see if there will be some updates :)

PR #293, opened 2017-02-21
Renamed namespace from Resource\Api to Model

MichalHanajik:
Renamed all sources of **Resource\Api** to **Model** in all files as mentioned in #289


MichalHanajik:
Of course I can, I was not sure if this is already as part of Models in this case.
Should I update already all models final from #288  ?

Nyholm:
Sure. Let's. But do not solve any more issues in this PR. Fixing #288 and #289 in this PR is enough.

Nyholm:
Also, when making "all models final" you have to be aware that some models cannot be marked final because another model extends that one.

MichalHanajik:
Added final for classes. Should be alright.

Nyholm:
Awesome. Just fix the minor style issue and I'll be happy to merge

MichalHanajik:
Fixed :)

Nyholm:
Thanks

PR #294, opened 2017-02-21
Fix integration tests for domain API

z38:
The integration tests for the Domain API were failing because non-GET requests did not work properly and in cases when an error occured, a generic `ErrorResponse` was actually returned.

### DIFF
--- a/src/Mailgun/Api/HttpApi.php
+++ b/src/Mailgun/Api/HttpApi.php
@@ -129,18 +129,16 @@ abstract class HttpApi
     }
 
     /**
-     * Send a POST request with JSON-encoded parameters.
+     * Send a POST request with parameters.
      *
      * @param string $path           Request path
-     * @param array  $parameters     POST parameters to be JSON encoded
+     * @param array  $parameters     POST parameters
      * @param array  $requestHeaders Request headers
      *
      * @return ResponseInterface
      */
     protected function httpPost($path, array $parameters = [], array $requestHeaders = [])
     {
-        $requestHeaders['Content-Type'] = 'application/x-www-form-urlencoded';
-
         return $this->httpPostRaw($path, $this->createRequestBody($parameters), $requestHeaders);
     }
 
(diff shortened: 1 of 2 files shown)


tests/Integration/DomainApiTest.php
```diff
@@ -77,7 +77,7 @@ public function testRemoveDomainNoExist()
 
         $ret = $mg->domains()->delete('example.notareal.tld');
         $this->assertNotNull($ret);
-        $this->assertInstanceOf(DeleteResponse::class, $ret);
+        $this->assertInstanceOf(ErrorResponse::class, $ret);
```
Nyholm: Is this really needed? Maybe a separate PR?

tests/Integration/DomainApiTest.php
```diff
@@ -318,7 +318,7 @@ public function testRemoveCredentialNoExist()
             $login
         );
         $this->assertNotNull($ret);
-        $this->assertInstanceOf(DeleteCredentialResponse::class, $ret);
+        $this->assertInstanceOf(ErrorResponse::class, $ret);
```
Nyholm: Maybe a separate PR?

src/Mailgun/Api/HttpApi.php
```diff
@@ -182,14 +182,27 @@ protected function httpDelete($path, array $parameters = [], array $requestHeade
...
+            }
+            foreach ($values as $value) {
+                $resources[] = [
+                    'name' => $key,
+                    'content' => $value,
+                ];
+            }
+        }
+
+        return $resources;
```
Nyholm: Good.

z38:
@Nyholm I'm not quite sure how to test for server errors. Should we just check that the appropriate exception is thrown (one of those in `HttpApi::handleErrors(...)`) or should we explicitly decode and check the returned message (with the help of #294)?

Nyholm:
We should only check that the correct exception is thrown.

z38:
@Nyholm The patch is now rebased against `master`. I manually verified that the integrations tests were successful (not covered by Travis).

PR #295, opened 2017-02-28
Added deprecation notices

Nyholm:


Nyholm:
Thank you for the review

dimsav:
Hello,

the Mailgun::get() method seems to be deprecated but using it seems to be the only way of using the `address/validate` endpoint.

I couldn't find any alternative in the docs or anywhere in the code. Can you please clarify?

Thank you

@Nyholm @DavidGarciaCat

DavidGarciaCat:
Hi @dimsav 

We have a WIP PR for EMail validation here https://github.com/mailgun/mailgun-php/pull/427, if this is what you mean.

Please don't hesitate to review the **Roadmap to v3** https://github.com/mailgun/mailgun-php/issues/442 and add new issues if you think we're missing something, so we can check if we need to add before to release the next major version.

Locking this conversation now due we have other issues to track this case.

Cheers,

PR #300, opened 2017-03-13
Fix setting Reply-To multiple times

### LINKED ISSUE #299: Can not override h:reply-to using setReplyToAddress
Because it throws an exception "array_push() expects parameter 1 to be array, string given" on 114 line in Mailgun\Messages\MessageBuilder.

z38:
Fixes #299

### DIFF
--- a/src/Mailgun/Messages/MessageBuilder.php
+++ b/src/Mailgun/Messages/MessageBuilder.php
@@ -117,10 +117,10 @@ class MessageBuilder
     {
         $compiledAddress = $this->parseAddress($address, $variables);
 
-        if (isset($this->message[$headerName])) {
-            array_push($this->message[$headerName], $compiledAddress);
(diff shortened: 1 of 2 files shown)


src/Mailgun/Messages/MessageBuilder.php
```diff
@@ -117,10 +117,10 @@ protected function addRecipient($headerName, $address, $variables)
     {
         $compiledAddress = $this->parseAddress($address, $variables);
 
-        if (isset($this->message[$headerName])) {
-            array_push($this->message[$headerName], $compiledAddress);
-        } elseif ($headerName == 'h:reply-to') {
+        if ($headerName == 'h:reply-to') {
```
DavidGarciaCat: I would say to compare strictly `===`

z38: :+1: Done

PR #304, opened 2017-03-25
Added factory methods

Nyholm:
It make it easier to create an instance of the API


Nyholm:
Thank you for the review.

Issue #281, opened 2017-02-15
Unable to create a new route


z38:
I tried to create a new route using the new API:
```php
$mailgun = new Mailgun\Mailgun('...');
$response = $mailgun->routes()->create('catch_all()', ['forward("mailbox@myapp.com")'], 'example', 100);
var_dump($response);
```

However, an error is returned (even though the expression is sent):
```text
object(Mailgun\Resource\Api\ErrorResponse)#23 (2) {
  ["message":"Mailgun\Resource\Api\ErrorResponse":private]=>
  string(21) "Expression is missing"
  ["error":"Mailgun\Resource\Api\ErrorResponse":private]=>
  NULL
}
```

Nyholm:
Closed by #313

PR #306, opened 2017-03-25
Added test for Route::Create

Nyholm:
Related to #281


Nyholm:
This PR just adds some tests

Issue #287, opened 2017-02-18
Improve safeDeserialize


Nyholm:
Make sure to catch all common errors and throw exceptions. 

Related to #276

PR #307, opened 2017-03-25
Improving safe hydrate

Nyholm:
I renamed in "HydrateResponse" and I moved out the error checking to a separate function.

### DIFF
--- a/src/Mailgun/Api/HttpApi.php
+++ b/src/Mailgun/Api/HttpApi.php
@@ -59,7 +59,7 @@ abstract class HttpApi
      * @param ResponseInterface $response
      * @param string            $class
      *
-     * @return mixed|array|ResponseInterface
+     * @return mixed|ResponseInterface
      *
      * @throws \Exception
      */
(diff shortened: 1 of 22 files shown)


Nyholm:
Everything that does not respond with 200 or 201 will be treated as an error by default

src/Mailgun/Api/HttpApi.php
```diff
@@ -56,38 +56,58 @@ public function __construct(HttpClient $httpClient, RequestBuilder $requestBuild
...
-     * Attempts to safely deserialize the response into the given class.
-     * If the HTTP return code != 200, deserializes into SimpleResponse::class
-     * to contain the error message and any other information provided.
-     *
      * @param ResponseInterface $response
-     * @param string            $className
+     * @param string            $class
      *
-     * @throws HttpClientException
+     * @return mixed|array|ResponseInterface
```
DavidGarciaCat: Are you sure we need to change the `@return` adding `mixed`? That looks wrong to me, if we don't know the type of content we're returning

Nyholm: Technically. We do not know at all what to return. Since we do not know what the Hydrator return. Maybe it should be `mixed|ResponseInterface` since that is the only thing we are sure at.

Nyholm:
This will fix #287

PR #308, opened 2017-03-26
Renamed Routes to Route

Nyholm:
Class names should be in singlular


Nyholm:
Thank you. You are correct. I missed those.

PR #313, opened 2017-03-26
Do not use json request body

Nyholm:
Build a query encoded request strings and set proper headers.

This will fix issue #281


Nyholm:
Sure, the API could support json, but we need proper headers. 

Also, when we use multipart stream we are not using json, why not always use application/x-www-form-urlencoded?

z38:
Thanks for taking a look at this. This PR will reintroduce the problems we are trying to fix in #279. `http_build_query()` uses brackets when multiple parameters have the same key. In #294 I proposed an alternative solution which uses multipart requests (which do not use brackets in turn).

Nyholm:
oh, I have to review that PR then. Just a sec

PR #314, opened 2017-03-26
Fix new constructor

Nyholm:
Make sure we use factories

TODO
- [x] Add docs


Nyholm:
Thank you for merging.

PR #325, opened 2017-04-07
Implemented a way to verify webhook signature

### LINKED ISSUE #321: Unable to verify/handle webhooks with the new API
`Mailgun::verifyWebhookSignature()` is deprecated and will be removed in 3.0. `Mailgun\Api\Webhook` does not offer this functionality yet. It might be worth investigating if the received events should be parsed and wrapped in models too.

Nyholm:
This will fix #321 
I just moved the function to the Webhook API

### DIFF
--- a/src/Mailgun/Mailgun.php
+++ b/src/Mailgun/Mailgun.php
@@ -84,20 +84,20 @@ class Mailgun
     }
 
     /**
-     * @param HttpClientConfigurator $httpClientConfigurator
+     * @param HttpClientConfigurator $configurator
      * @param Hydrator|null          $hydrator
      * @param RequestBuilder|null    $requestBuilder
      *
      * @return Mailgun
      */
     public static function configure(
-        HttpClientConfigurator $httpClientConfigurator,
+        HttpClientConfigurator $configurator,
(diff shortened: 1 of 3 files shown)


src/Mailgun/Mailgun.php
```diff
@@ -84,20 +84,20 @@ public function __construct(
...
-        HttpClientConfigurator $httpClientConfigurator,
+        HttpClientConfigurator $configurator,
         Hydrator $hydrator = null,
         RequestBuilder $requestBuilder = null
     ) {
-        $httpClient = $httpClientConfigurator->createConfiguredClient();
+        $httpClient = $configurator->createConfiguredClient();
 
-        return new self(null, $httpClient, 'api.mailgun.net', $hydrator, $requestBuilder);
+        return new self($configurator->getApiKey(), $httpClient, 'api.mailgun.net', $hydrator, $requestBuilder);
```
DavidGarciaCat: Why this change? It was not here until now, and I don't see calls to this static method (or I am too tired).

Nyholm: The webhook API need the API key to be able to verify the signature.

PR #328, opened 2017-04-07
Added documentation for the API classes

### LINKED ISSUE #273: Document how to set attachment name
See https://github.com/mailgun/mailgun-php/issues/21

Nyholm:
This will fix #273

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,11 @@ $mg->message()->send('example.com', [
 ]);
 ```
 
+### All usage examples
+
+You find more detailed documentation at in [/doc](doc/index.md) and on 
+[https://documentation.mailgun.com](https://documentation.mailgun.com/api_reference.html).
+
 ### Response
 
 The results of a API call is, by default, a domain object. This will make it easy
--- a/README.md
+++ b/README.md
@@ -130,12 +130,12 @@ $configurator->setDebug(true);
 $mg = Mailgun::configure($configurator);
 
 # Now, compose and send your message.
-$mg->sendMessage('example.com', [
+$mg->messages()->sendMessage('example.com', [
   'from'    => 'bob@example.com', 
   'to'      => 'sally@example.com', 
   'subject' => 'The PHP SDK is awesome!', 
-  'text'    => 'It is so simple to send a message.']
-);
+  'text'    => 'It is so simple to send a message.'
+]);
 ```
 ### Additional Info
 
--- a/doc/attachments.md
+++ b/doc/attachments.md
@@ -1,13 +1,10 @@
-#Attachments
+# Attachments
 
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -130,12 +135,12 @@ $configurator->setDebug(true);
 $mg = Mailgun::configure($configurator);
 
 # Now, compose and send your message.
-$mg->sendMessage('example.com', [
+$mg->messages()->sendMessage('example.com', [
```
DavidGarciaCat: I would like to consider an update here: if we're calling `->messages()` then - obviously - I will work with messages, so maybe makes sense just `->send()`

doc/attachments.md
```diff
@@ -0,0 +1,53 @@
...
+
+You may attach a file from memory or by a file path.
+
+## From file path
+
+```php
+// Some how load the file to memory
+$binaryFile = '[Binary data]';
+
+$mg->message()->send('example.com', [
```
DavidGarciaCat: Or the previous comment is showing an error in the docs 😜

README.md
Nyholm: You are correct. The method is called `send`

doc/attachments.md
Nyholm: Thanks!

doc/attachments.md
```diff
@@ -0,0 +1,53 @@
...
+    ['fileContent'=>$binaryFile, 'filename'=>'test.jpg']
+  ]
+]);
+```
+
+## Inline attachments
+
+```php
+// Some how load the file to memory
+$binaryFile = '[Binary data]';
```
pirogoeth: I don't think we need this here since this example is showing the `'filePath'` method.

doc/index.md
```diff
@@ -0,0 +1,311 @@
...
+
+Other relevant documentation pages might be: 
+ 
+* [Attachments](attachments.md)
+* [Pagination](pagination.md)
+* [Message Builder](src/Mailgun/Messages/README.md) (Legacy code)
+* [Batch Message](src/Mailgun/Messages/README.md) (Legacy code)
+* [Opt-In Handler](src/Mailgun/Lists/README.md)  (Legacy code)
+
+##Domain API
```
pirogoeth: You need to insert a space between the `#` and the title text or the markdown will not render correctly. Otherwise, awesome!

doc/attachments.md
Nyholm: Thank you

doc/index.md
Nyholm: Yeah, you are correct. It is a bug in my local "Github markdown renderer". Thanks!

PR #331, opened 2017-04-07
Removed old "shared host install".

Nyholm:
It is 2017. You should not use FTP servers anymore. Composer just turned 6 years old. It is time we drop this from our documentation.


Nyholm:
Thank you for merging.

PR #332, opened 2017-04-07
Fixing broken suppression API

Nyholm:


Nyholm:
This diff is not fair. I just moved some classes around to create "Bounce" and ShowResponse.

PR #335, opened 2017-04-08
[RFC] Re-enable integration tests

z38:
To ensure that the integration tests don't fall behind the current development, it might make sense to enable them again. The testsuite only covers a small subset of the whole API so updating it won't cause much work and ensures that this library is working (and continues to).


Nyholm:
Would you mind rebasing this on master? I just merged #337

Nyholm:
I just ran it locally. Ill merge this now and deal with potential errors afterwards.

PR #339, opened 2017-04-08
Use latest version of guzzle/psr7

Nyholm:


Nyholm:
Without this the "test with lowest dependencies" would fail

PR #341, opened 2017-04-10
Support mime messages

Nyholm:
Make sure we support Mime. This is not a BC.. I'll explain in the comments.

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -97,7 +97,6 @@ $mailgun->messages()->send('example.com', $parameters);
 Below in an example how to create a Mime message with SwiftMailer.
 
 ```php
-
 $message = \Swift_Message::newInstance('Mail Subject');
 $message->setFrom(['from@exemple.com' => 'Example Inc']);
 $message->setTo(['user0gmail.com' => 'User 0', 'user1@hotmail.com' => 'User 1']);
@@ -174,7 +173,6 @@ $mailgun->stats()->total('example.com');
 ```php
 $mailgun->stats()->all('example.com');
 ```
-
 
 ## Suppression API
 
--- a/doc/index.md
+++ b/doc/index.md
@@ -92,6 +92,27 @@ $parameters = [
 ];
 $mailgun->messages()->send('example.com', $parameters);
 ```
+#### Send a message with Mime
+
+Below in an example how to create a Mime message with SwiftMailer.
+
+```php
+
+$message = \Swift_Message::newInstance('Mail Subject');
+$message->setFrom(['from@exemple.com' => 'Example Inc']);
+$message->setTo(['user0gmail.com' => 'User 0', 'user1@hotmail.com' => 'User 1']);
+// $message->setBcc('admin@example.com'); Do not do this, BCC will be visible for all receipients if you do. 
+$message->setCc('invoice@example.com');
+
+$messageBody = 'Look at the <b>fancy</b> HTML body.';
+$message->setBody($messageBody, 'text/html');
+
+// We need all "tos". Incluce the BCC here.
+$to = ['admin@example.com', 'user0gmail.com', 'user1@hotmail.com', 'invoice@example.com']
+
+// Send the message
+$mailgun->messages()->sendMime('example.com', $to, $message->toString());
+```
 
 #### Show a stored message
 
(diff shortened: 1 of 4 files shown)


src/Mailgun/Api/Message.php
```diff
@@ -30,18 +30,21 @@ public function send($domain, array $params)
...
+            $meme = '.mime';
+        }
+
         $postDataMultipart = [];
         $fields = ['message', 'attachment', 'inline'];
         foreach ($fields as $fieldName) {
             if (!isset($params[$fieldName])) {
                 continue;
             }
-            if (!is_array($params[$fieldName])) {
```
Nyholm: This can be removed since if this is not an array, we would get a type error calling prepareFile.

Nyholm:
Nooo!
Fetching `https://api.mailgun.net/non.existing.uri/1/2/3` will no longer result in 404 (which it should). Mailgun has decided to always reply with 200... That is wrong.. hm.. removing that tests now. (It was not a good test to begin with)

src/Mailgun/Api/Message.php
```diff
@@ -20,59 +20,74 @@
...
      * @return SendResponse
      */
     public function send($domain, array $params)
     {
+        Assert::string($domain);
         Assert::notEmpty($domain);
         Assert::notEmpty($params);
 
         $postDataMultipart = [];
-        $fields = ['message', 'attachment', 'inline'];
```
Nyholm: This can be removed since it had no effect.

Nyholm:
I've added tests and docs. This PR is ready for review.

doc/index.md
```diff
@@ -92,6 +92,27 @@ $parameters = [
 ];
 $mailgun->messages()->send('example.com', $parameters);
 ```
+#### Send a message with Mime
+
+Below in an example how to create a Mime message with SwiftMailer.
+
+```php
+
```
DavidGarciaCat: No need to keep this empty line

src/Mailgun/Api/Message.php
```diff
@@ -20,59 +20,74 @@
...
-
+        $postDataMultipart = array_merge($this->prepareMultipartParameters($params), $postDataMultipart);
         $response = $this->httpPostRaw(sprintf('/v3/%s/messages', $domain), $postDataMultipart);
 
         return $this->hydrateResponse($response, SendResponse::class);
     }
 
     /**
+     * @param string $domain
+     * @param array  $to      with all you send emails to. Including bcc and cc
```
DavidGarciaCat: I think the `@param` comment here is confusing, specially because `@param` name is `$to` and comment says `including BCC and CC`. If this parameter expects **all** recipients then I suggest to rename the parameter name

DavidGarciaCat: Also, if we're adding now the `@param` comments then will be nice to have them for each parameter, not just 2 of 4 😄

src/Mailgun/Api/Message.php
```diff
@@ -129,4 +144,33 @@ private function prepareFile($fieldName, array $filePath)
...
+     * @param array $params
+     *
+     * @return array
+     */
+    private function prepareMultipartParameters(array $params)
+    {
+        $postDataMultipart = [];
+        foreach ($params as $key => $value) {
+            if (is_array($value)) {
+                foreach ($value as $subValue) {
```
DavidGarciaCat: Can we update this, like a recursive call? Something like this?
```php
if (is_array($value)) {
    $postDataMultipart  = $this->prepareMultipartParameters($value);
}
```

src/Mailgun/Api/Message.php
Nyholm: I named it "to" to reflect the API docs. But you are correct. I reamed it now. 

About the comments, There is no need to comment "$domain"... it is used on all endpoints. The "$param", sure that may need a comment, or we could just trust people to read the API doc like we do an other endpoints.

src/Mailgun/Api/Message.php
Nyholm: hm.. We can improve this. But recursion is wrong because we do not allow a tree with depth more that 2.

Nyholm:
Thank you for the review. Im quite happy with the new `prepareMultipartParameters`. What do you think?

DavidGarciaCat:
Looks fine to me - @pirogoeth thoughts?

PR #342, opened 2017-04-11
Do not skip route test

Nyholm:


Nyholm:
Test are failing because of https://github.com/mailgun/mailgun-php/pull/341#issuecomment-293074670

PR #347, opened 2017-05-02
Show that the constructor is deprecated

Nyholm:


src/Mailgun/Mailgun.php
```diff
@@ -67,8 +67,10 @@ class Mailgun
      * @param string              $apiEndpoint
      * @param Hydrator|null       $hydrator
      * @param RequestBuilder|null $requestBuilder
+     *
+     * @internal Use Mailgun::configure or Mailgun::create instead.
      */
-    public function __construct(
+    public /* private */ function __construct(
```
DavidGarciaCat: Please, do not comment code. If you think it is no longer useful, just drop it. If we need to restore the existing code then we can check the file history.

Nyholm: I did not change the constructor. This PR is only a doc fix. 

I added this comment to show that it should be treated as "private". This not code. It is a way to show that the constructor is private without breaking BC.

PR #349, opened 2017-05-05
fix: named addresses must be surrounded by double quotes

AlexDpy:
Due to your recent API update, named addresses with special chars (like parentheses) are now rejected if they are surrounded by simple quotes
They must be surrounded by double quotes

'Whoever (SomeCompany)' <some@address.com> is no longer valid and will be rejected by your API
"Whoever (SomeCompany)" <some@address.com> is valid and will be OK

### DIFF
--- a/src/Mailgun/Messages/MessageBuilder.php
+++ b/src/Mailgun/Messages/MessageBuilder.php
@@ -102,7 +102,7 @@ class MessageBuilder
         }
         $fullName = $this->getFullName($variables);
         if ($fullName != null) {
-            return "'$fullName' <$address>";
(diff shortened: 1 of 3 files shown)


AlexDpy:
Ok I've just updated the PR with `sprintf`

Concerning the extra conditional you're talking about, it seems that it's already effective [here](https://github.com/mailgun/mailgun-php/blob/master/src/Mailgun/Messages/MessageBuilder.php#L104)

Nyholm:
Great. Thank you for this PR. 
Also, excellent review!

Nyholm:
Great. Thank you for this PR. 
Also, excellent review!

PR #356, opened 2017-05-21
Use stable version of multipart stream builder

Nyholm:
There were no changes from 0.2.0 to 1.0.0


Nyholm:
Old projects (previous versions) will still work. If you are having a conflict (ie two libraries requiring multipart-stream-builder), one should update the other library to 1.0.0. of the multipart-stream-builder .

PR #358, opened 2017-05-22
Few test corrections to preview where it goes.

TemirkhanN:


tests/RequestBuilderTest.php
```diff
@@ -9,37 +9,151 @@
 
 namespace Mailgun\Tests;
 
+use Http\Message\MultipartStream\MultipartStreamBuilder;
+use Http\Message\RequestFactory;
 use Mailgun\RequestBuilder;
+use PHPUnit\Framework\TestCase;
```
Nyholm: This will require phpunit 4.8.25 or later.

tests/RequestBuilderTest.php
```diff
@@ -9,37 +9,151 @@
...
+use PHPUnit_Framework_MockObject_MockObject as MockObject;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\StreamInterface;
 
-class RequestBuilderTest extends \PHPUnit_Framework_TestCase
+/**
+ * TODO request factory instantiate should be tested somehow. For reason something seems to be wrong with realization
```
Nyholm: This comment should be resolved or removed.

tests/RequestBuilderTest.php
TemirkhanN: But there is
https://github.com/mailgun/mailgun-php/blob/master/composer.json#L14

tests/RequestBuilderTest.php
TemirkhanN: The only way to write test without changing logic in base class is to keep old output tests. But I think It wouldn't be correct. What do you think?

tests/RequestBuilderTest.php
Nyholm: Yes, but you will fail the "prefer-lowest" tests because there is no patch version specified. 

https://travis-ci.org/mailgun/mailgun-php/builds/234785059?utm_source=github_status&utm_medium=notification

tests/RequestBuilderTest.php
```diff
@@ -12,15 +12,12 @@
...
 use PHPUnit_Framework_MockObject_MockObject as MockObject;
+use PHPUnit_Framework_TestCase;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\StreamInterface;
 
-/**
- * TODO request factory instantiate should be tested somehow. For reason something seems to be wrong with realization
- */
-class RequestBuilderTest extends TestCase
+class RequestBuilderTest extends PHPUnit_Framework_TestCase
```
Nyholm: This also works, but there is no PHPUnit 6 support. We'll fix that later.

PR #362, opened 2017-06-06
Fix batch message undefined property error

TheoKouzelis:
While testing Batch Messages against Postbin I got an error of "Undefined property: stdClass::$id" because post bin doesn't return an id in it's response message. So I wrote this PR to get Batch Messages to check for id's in responses.

### DIFF
--- a/tests/Mock/Connection/PostBinBroker.php
+++ b/tests/Mock/Connection/PostBinBroker.php
@@ -1,0 +1,15 @@
+<?php
+
+/*
+ * Copyright (C) 2013-2016 Mailgun
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+namespace Mailgun\Tests\Mock\Connection;
+
+class PostBinBroker extends TestBroker
+{
+    protected $responseData = '{"message": "Some JSON Response Data"}';
+}
--- a/tests/Mock/Connection/TestBroker.php
+++ b/tests/Mock/Connection/TestBroker.php
@@ -21,6 +21,8 @@ class TestBroker extends RestClient
 
     protected $apiEndpoint;
 
+    protected $responseData = '{"message": "Some JSON Response Data", "id": "1234"}';
+
     public function __construct($apiKey = null, $apiHost = 'api.mailgun.net', $apiVersion = 'v3')
     {
         $this->apiKey = $apiKey;
@@ -52,7 +54,7 @@ class TestBroker extends RestClient
         if ($httpResponseCode === 200) {
             $result = new \stdClass();
             $result->http_response_body = new \stdClass();
-            $jsonResponseData = json_decode('{"message": "Some JSON Response Data", "id": "1234"}');
+            $jsonResponseData = json_decode($this->responseData);
             foreach ($jsonResponseData as $key => $value) {
                 $result->http_response_body->$key = $value;
             }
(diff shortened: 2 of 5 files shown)


tests/Mock/Connection/TestBroker.php
```diff
@@ -21,6 +21,8 @@ class TestBroker extends RestClient
 
     protected $apiEndpoint;
 
+    protected $responseData = '{"message": "Some JSON Response Data", "id": "1234"}';
```
DavidGarciaCat: Personally, if you are not going to reuse this property on other methods, there is no need to define it as shared/protected property, and its content can be moved directly to line 57, inside the `json_decode` function.

tests/Mock/Connection/PostBinBroker.php
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+namespace Mailgun\Tests\Mock\Connection;
+
+class PostBinBroker extends TestBroker
+{
+    protected $responseData = '{"message": "Some JSON Response Data"}';
```
DavidGarciaCat: Please refer to `tests/Mock/Connection/TestBroker.php` file comment.
This is the same case.

tests/Mock/Connection/TestBroker.php
TheoKouzelis: Hi @DavidGarciaCat, Thanks for the review. I moved it to a property so I could extend the class and override the property to mock different response messages. Is there a better way to achieve this?

tests/Mock/Connection/PostBinBroker.php
TheoKouzelis: @DavidGarciaCat This is where I am overriding the property to mock the response from PostBin

z38:
I'm not sure whether this is the correct way to approach this. The purpose of Postbin is not to return valid data but to see what's in the POST request body (i.e. useful when testing webhooks). Most requests of this library will fail when used with Postbin. The methods are all built to be tested against a real (or at least similar looking) server. With this change whenever the real backend returns an malformed message, it's just ignored.

tests/Mock/Connection/TestBroker.php
TheoKouzelis: Maybe a setter method would make more sense?

TheoKouzelis:
As a user, I was a bit confused when I was testing my app against Postbin and was getting PHP errors. I thought this PR would help save people time look for potential bugs in their apps. Maybe getting Postbin to return an ID in its response would be a better approach?

z38:
@TheoKouzelis I can totally understand the confusion. IMHO we should clearly state that Postbin can just be used to see what would have been sent to Mailgun but not to test whether a method would actually succeed.

tests/Mock/Connection/TestBroker.php
DavidGarciaCat: Hi @TheoKouzelis, IMHO if you're going to override it to test each possible case, then does not make sense (to me) to define it as protected in the parent class. Based on this:
1. A valid option is to remove the parent `protected` property and set it as `private` on the child property.
2. As described above, if we are not going to use it in more than one test (AKA if we don't need to reuse it, or use it more than once) then does not make sense to define it at Class level due it is just used inside one of the functions

Additionally, even when some times `getters` and `setters` are useful, abuse of these methods is not a good idea. You can [read this blog post](http://www.javaworld.com/article/2073723/core-java/why-getter-and-setter-methods-are-evil.html) if you want to know why.

TheoKouzelis:
OK cool, thanks for reviewing the PR.

tests/Mock/Connection/TestBroker.php
TheoKouzelis: Thanks, I will keep that in mind for the future.

DavidGarciaCat:
More than 2 and a half month - are we waiting for something here?

Nyholm:
Closing due inactivity and this code is deprecated.

PR #365, opened 2017-06-20
Fix DNS record validation

maximzasorin:


Nyholm:
Thank you.

Nyholm:
I tagged a new patch version with your fix.

PR #371, opened 2017-06-23
Expose response code from HttpClientException

pirogoeth:
A complaint I received is that it was no longer easy to determine, for example, if a bounce a user is trying to `GET` is not found or if another API error occurred. This should help to ease that issue.

### DIFF
--- a/src/Mailgun/Exception/HttpClientException.php
+++ b/src/Mailgun/Exception/HttpClientException.php
@@ -28,6 +28,11 @@ final class HttpClientException extends \RuntimeException implements Exception
     private $responseBody;
 
     /**
+     * @var int
+     */
+    private $responseCode;
+
+    /**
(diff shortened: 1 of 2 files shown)


src/Mailgun/Exception/HttpClientException.php
```diff
@@ -38,6 +43,7 @@ public function __construct($message, $code, ResponseInterface $response = null)
 
         if ($response) {
             $this->response = $response;
+            $this->responseCode = $code;
```
Nyholm: I think you mean `$response->getStatusCode()`

DavidGarciaCat: agree

pirogoeth: Fixed, thanks!

PR #379, opened 2017-08-04
Enable parsing webhook POST data into ShowResponse

michbr:
Change the ShowResponse::create() method to support the data format from a webhook POST.

### DIFF
--- a/src/Mailgun/Model/Message/ShowResponse.php
+++ b/src/Mailgun/Model/Message/ShowResponse.php
@@ -142,7 +142,8 @@ class ShowResponse implements ApiResponse
             $response->setMessageUrl($data['message-url']);
         }
         if (isset($data['message-headers'])) {
-            $response->setMessageHeaders($data['message-headers']);
+            $headers = is_string($data['message-headers']) ? json_decode($data['message-headers']) : $data['message-headers'];
+            $response->setMessageHeaders($headers);
         }
         if (isset($data['recipient'])) {
             $response->setRecipient($data['recipient']);


src/Mailgun/Model/Message/ShowResponse.php
```diff
@@ -142,7 +142,8 @@ public static function create(array $data)
             $response->setMessageUrl($data['message-url']);
         }
         if (isset($data['message-headers'])) {
-            $response->setMessageHeaders($data['message-headers']);
+            $headers = is_string($data['message-headers']) ? json_decode($data['message-headers']) : $data['message-headers'];
```
Nyholm: Will `$data['message-headers']` ever be a string?

Could you provide a test for this?

michbr: $data['message-headers'] is always a string in the case that the message was received as a POST from the routes configured at [https://app.mailgun.com/app/routes](https://app.mailgun.com/app/routes).

DavidGarciaCat: Hey there,

I don't believe this will work, based on a quick look on API docs, I believe a test is mandatory here.

https://documentation.mailgun.com/en/latest/api-routes.html#examples

screen shot 2017-08-05 at 00 34 20

michbr: As I said, I will provide a test...when I get around to it. In the meantime, as a work around, I am already doing the following:
`$data= <object I got from the POST>`
$data['message-headers'] = json_decode($data['message-headers']);
ShowResponse::create($data);`

And that's working great. So it's working for me...

michbr:
As it turns out, this doesn't work. Apparently a retrieved message has different data and formatting than a message forwarded by a route. Furthermore, when a route "notifies" of a storage event, it sends the same data as when it forwards the message itself. 

The server side behavior is too ridiculous for the php sdk to make up for it, unfortunately. Ideally a forwarded message would be the same as a retrieved message, and a route notify would send a proper Event object.

DavidGarciaCat:
That's sad, indeed. Thanks for keep us posted.

PR #384, opened 2017-08-23
Fixed stats retrieval

deguif:
Currently when using the `stats->total()` method, it always returns an empty array for stats.

- There was an issue in the code, with a wrong isset check.
- There was a second issue on `TotalResponseItem` as some constructor parameters are typed as array but in some case, the null value is passed instead.

There is also a third issue (I don't take time to resolve as it's more difficult), as we can't specified a list of stats events to retrieve from the api as the query is not formatted as expected, which makes the method only usable with one event, not several.

Mailgun expects `/v3/xxx/stats/total?duration=100d&event=delivered&event=failed`

Instead the mailgun-php library formats it this way: `/v3/xxx/stats/total?duration=100d&event%5B0%5D=delivered&event%5B1%5D=failed`

As fixing this requires updating the `httpGet` method from the `HttpApi` class and not using the `http_build_query` function, I don't know what is the best way to fix it.

There is also a fourth issue (that could probably be solved here too, just need to confirm if wanted), as  currently we can only retrieve the failed, delivered and accepted stats event, but there are a lot more of events, see http://mailgun-documentation.readthedocs.io/en/latest/api-stats.html#event-types

### DIFF
--- a/src/Mailgun/Model/Stats/TotalResponse.php
+++ b/src/Mailgun/Model/Stats/TotalResponse.php
@@ -58,7 +58,7 @@ final class TotalResponse implements ApiResponse
     public static function create(array $data)
     {
         $stats = [];
-        if (isset($data['status'])) {
+        if (isset($data['stats'])) {
             foreach ($data['stats'] as $s) {
                 $stats[] = TotalResponseItem::create($s);
             }
--- a/src/Mailgun/Model/Stats/TotalResponseItem.php
+++ b/src/Mailgun/Model/Stats/TotalResponseItem.php
@@ -43,9 +43,9 @@ class TotalResponseItem
     {
         return new self(
             isset($data['time']) ? new \DateTime($data['time']) : null,
-            isset($data['accepted']) ? $data['accepted'] : null,
-            isset($data['delivered']) ? $data['delivered'] : null,
-            isset($data['failed']) ? $data['failed'] : null
+            isset($data['accepted']) ? $data['accepted'] : [],
+            isset($data['delivered']) ? $data['delivered'] : [],
+            isset($data['failed']) ? $data['failed'] : []
         );
     }
 
--- a/tests/Api/StatsTest.php
+++ b/tests/Api/StatsTest.php
@@ -10,6 +10,9 @@
 namespace Mailgun\Tests\Api;
 
 use GuzzleHttp\Psr7\Response;
+use Mailgun\Hydrator\ModelHydrator;
+use Mailgun\Model\Stats\TotalResponse;
+use Mailgun\Model\Stats\TotalResponseItem;
 
 /**
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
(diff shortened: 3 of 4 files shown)


deguif:
any news on this ?
ping @contributors

deguif:
Yeah sure.
Also, maybe I should use normal `array` syntax instead of `[]`, I just saw I pushed my code with this syntax?

DavidGarciaCat:
From my POV use `[]` is not a problem due the minimum version of PHP is 5.5 and it is a valid syntax.

deguif:
Just added tests, sorry for the delay, it was hard to find time to work on it.

src/Mailgun/Model/Stats/TotalResponseItem.php
```diff
@@ -43,9 +43,9 @@ public static function create(array $data)
     {
         return new self(
             isset($data['time']) ? new \DateTime($data['time']) : null,
-            isset($data['accepted']) ? $data['accepted'] : null,
-            isset($data['delivered']) ? $data['delivered'] : null,
-            isset($data['failed']) ? $data['failed'] : null
+            isset($data['accepted']) ? $data['accepted'] : [],
+            isset($data['delivered']) ? $data['delivered'] : [],
+            isset($data['failed']) ? $data['failed'] : []
```
DavidGarciaCat: Question: why to set `[]` instead of `null`? IMO if we don't have that array key, then `null` is fine.

src/Mailgun/Model/Stats/TotalResponse.php
```diff
@@ -58,7 +58,7 @@ private function __construct(\DateTime $start, \DateTime $end, $resolution, arra
     public static function create(array $data)
     {
         $stats = [];
-        if (isset($data['status'])) {
+        if (isset($data['stats'])) {
```
DavidGarciaCat: Nice catch!

src/Mailgun/Model/Stats/TotalResponseItem.php
deguif: The `TotalResponseItem` class requires array parameters in its constructor.
Without this there were errors as null was not accepted as a suitable constructor parameter.

This PR fixes two bugs ;)

DavidGarciaCat: That makes sense, however it means one more thing:

What about another update? Instead of:

```php
private function __construct(
    \DateTime $time,
    array $accepted,
    array $delivered,
    array $failed
)
```

the Constructor could be this:

```php
private function __construct(
    \DateTime $time,
    array $accepted = null,
    array $delivered = null,
    array $failed = null
)
```

So in this way we can think about the option that a given array key could not be passed.

DavidGarciaCat: In fact, this returns `null` if the keys are not provided:
https://github.com/mailgun/mailgun-php/blob/master/src/Mailgun/Model/Stats/TotalResponseItem.php#L42-L50

deguif: Allowing null values will require users to check if return of `TotalResponseItem::getAccepted()`, `TotalResponseItem::getDelivered()`, `..` is null before  iterating on stats items.

TRANSVESTIT31:
I am boys 32 year pasiv i heve shoot porno video porno foto

src/Mailgun/Model/Stats/TotalResponseItem.php
DavidGarciaCat: Good point, but before to go ahead with this, I would say to confirm - based on the official API docs - if this scenario could happen, because if not then we won't be providing a "pure" SDK that should provide the same behaviour that we could get running HTTP Requests to the API.

deguif:
Any chance to get this merged, now that tests were added.
ping @DavidGarciaCat @Nyholm

deguif:
Ping @Nyholm
Any news about this?

DavidGarciaCat:
Hey @pirogoeth don't know if @Nyholm is available these days, but feel free to share your thoughts

Nyholm:
Could you rebase on master?

deguif:
Yeah sure.
Now StyleCI is green

Nyholm:
Awesome. Thanks

Issue #343, opened 2017-04-11
Update StyleCi file headers


Nyholm:
They should read: 

```
Copyright (C) 2013 Mailgun

This software may be modified and distributed under the terms
of the MIT license. See the LICENSE file for details.
```

According to the copyright law Mailgun automatically have copyright for X (50?) years. We should only specify the year we first created this library. There is no need to have the current year.

PR #391, opened 2017-08-29
Bump year updating Copyright

shakaran:
This could be easily achieved in next years with the following bash command and changing the years:

`grep -l "2013-2016" | xargs sed -i 's/2013-2016/2013-2017/g'`

### DIFF
--- a/src/Mailgun/Api/Domain.php
+++ b/src/Mailgun/Api/Domain.php
@@ -1,7 +1,7 @@
 <?php
 
 /*
- * Copyright (C) 2013-2016 Mailgun
+ * Copyright (C) 2013-2017 Mailgun
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
--- a/src/Mailgun/Api/Event.php
+++ b/src/Mailgun/Api/Event.php
@@ -1,7 +1,7 @@
 <?php
 
 /*
- * Copyright (C) 2013-2016 Mailgun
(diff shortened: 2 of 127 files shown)


Nyholm:
Thank you. But this is controlled by style CI. I have a task for admin (@pirogoeth) to update the StyleCI settings for file headers: https://github.com/mailgun/mailgun-php/issues/343

shakaran:
@Nyholm since style CI is not supported with php-cs-fixer 2.x versions, StyleCI still will control this? Even if controlled or not, it needs to be updated properly, so I think that the ideal should be merge or at least upgrade .php_cs to use php-cs-fixer 2.x and update the headers

DavidGarciaCat:
Hey @shakaran - What @Nyholm means is this repository is inspected by Style CI when new commits are pushed, and the PHP headers that contain the years are part of the Style CI settings, so there's no need to apply this PR.

shakaran:
@DavidGarciaCat but since Style CI is not working(because not support for PHP-CS-Fixer 2.x branch) (or at least is not commiting and updating in old files), it never gets updated or I am wrong with that statment?

DavidGarciaCat:
Hey @shakaran - As @Nyholm said, issue https://github.com/mailgun/mailgun-php/issues/343 is related, and he has sent already a reminder to @pirogoeth via [Gitter](https://gitter.im/mailgun/mailgun-php), asking if he can deal with this request, so there is no need to keep this PR here.

Issue #390, opened 2017-08-29
Unsubscribe API responses different behaviour and missing params


shakaran:
I am trying two ways to get the list of unsuscribes for a domain, but I found diferent behaviours calling to different ways (which is pretty inconsistent):

If you check the unsubscribe API index in [API agnostic documentation](http://mailgun-documentation.readthedocs.io/en/latest/api-unsubscribes.html#unsubscribes):

1) It should allow two parameters: _skip_ and _limit_
2) The _limit_ says _100_ as maximum

But if you check the [source code in PHP API](
https://github.com/mailgun/mailgun-php/blob/master/src/Mailgun/Api/Suppression/Unsubscribe.php#L35) you will see:

1) It only allows _limit_ parameter (you cannot use the skip parameter)
2) The limit as default value is _100_, but, there is an assert between 1 and _10000_ ( _'Limit parameter must be between 1 and 10000'_ ), which confuse me even more.

But there are even more problems, which is is the way of how is returned the data from [Mailgun\Model\Suppression\Unsubscribe\IndexResponse](https://github.com/mailgun/mailgun-php/blob/master/src/Mailgun/Model/Suppression/Unsubscribe/IndexResponse.php)

For example:

```
use Mailgun\Model\Suppression\Unsubscribe\IndexResponse;

$result = $mailgun->suppressions()->unsubscribes()->index('example.com');

if($result instanceof IndexResponse)
{
    $total_unsubscribes = $result->getTotalCount(); // Missing param total_count
    $unsubscribes       = $result->getItems();
}
```

Instead if I call via:

```
$result = $mailgun->get('example.com/unsubscribes');

if(isset($result->http_response_code) && $result->http_response_code === 200)
{
    $total_unsubscribes = $result->http_response_body->total_count; // This record is available
    $unsubscribes       = $result->http_response_body->items;
}
```

Also since each item returned is an object of [Mailgun/Model/Suppression/Unsubscribe/Unsubscribe](https://github.com/mailgun/mailgun-php/blob/master/src/Mailgun/Model/Suppression/Unsubscribe/Unsubscribe.php), I notice which is missing the "id" fields which appears if you get only as json object via _$mailgun->get('example.com/unsubscribes');_ and the tag is returned as NULL instead "*" (which means all).

Conclusion:

[...]

PR #393, opened 2017-08-30
Implement getTotalCount and add doc descriptions

shakaran:
This should cover point 3 of issue #390

### DIFF
--- a/src/Mailgun/Model/Suppression/Unsubscribe/IndexResponse.php
+++ b/src/Mailgun/Model/Suppression/Unsubscribe/IndexResponse.php
@@ -35,7 +35,7 @@ final class IndexResponse implements ApiResponse, PagingProvider
      * @see Mailgun/Model/Suppression/Unsubscribe/Unsubscribe
      * @var integer
      */
-    private $total_count = 0;
+    private $totalCount = 0;
 
     /**
      * @see Mailgun/Model/Suppression/Unsubscribe/Unsubscribe
@@ -89,6 +89,8 @@ final class IndexResponse implements ApiResponse, PagingProvider
      */
     public function getTotalCount()
     {
-        return count($this->items);
+        $this->totalCount = count($this->items);
+        
+        return $this->totalCount;
     }
 }
--- a/src/Mailgun/Model/Suppression/Unsubscribe/IndexResponse.php
+++ b/src/Mailgun/Model/Suppression/Unsubscribe/IndexResponse.php
@@ -21,11 +21,25 @@ final class IndexResponse implements ApiResponse, PagingProvider
     use PaginationResponse;
 
     /**
+     * Array to store a list of Unsubscribe items from
+     * index response
+     * 
+     * @see Mailgun/Model/Suppression/Unsubscribe/Unsubscribe
      * @var Unsubscribe[]
      */
-    private $items;
+    private $items = [];
+    
+    /**
+     * Store the total number of Unsubscribe items
+     * 
+     * @see Mailgun/Model/Suppression/Unsubscribe/Unsubscribe
+     * @var integer
+     */
+    private $total_count = 0;
 
     /**
+     * @see Mailgun/Model/Suppression/Unsubscribe/Unsubscribe
+     * 
      * @param Unsubscribe[] $items
      * @param array         $paging
      */
@@ -36,6 +50,8 @@ final class IndexResponse implements ApiResponse, PagingProvider
     }
 
     /**
+     * Allow create the unsubscribe items with paging
+     * 
      * @param array $data
      *
      * @return IndexResponse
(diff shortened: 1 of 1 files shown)


src/Mailgun/Model/Suppression/Unsubscribe/IndexResponse.php
```diff
@@ -21,11 +21,25 @@
...
-    private $items;
+    private $items = [];
+    
+    /**
+     * Store the total number of Unsubscribe items
+     * 
+     * @see Mailgun/Model/Suppression/Unsubscribe/Unsubscribe
+     * @var integer
+     */
+    private $total_count = 0;
```
Nyholm: One should use camelCase

Nyholm: Also, do we ever read/write to this property?

src/Mailgun/Model/Suppression/Unsubscribe/IndexResponse.php
```diff
@@ -21,11 +21,25 @@
...
 
     /**
+     * Array to store a list of Unsubscribe items from
+     * index response
+     * 
+     * @see Mailgun/Model/Suppression/Unsubscribe/Unsubscribe
      * @var Unsubscribe[]
      */
-    private $items;
+    private $items = [];
```
DavidGarciaCat: IMHO we don't need this. Constructor already expects an array and the value will be set as an array.

src/Mailgun/Model/Suppression/Unsubscribe/IndexResponse.php
```diff
@@ -53,10 +69,28 @@ public static function create(array $data)
...
+    /**
+     * Get the total count of Unsusbscribe in index response
+     * 
+     * @see Mailgun/Model/Suppression/Unsubscribe/Unsubscribe
+     * 
+     * @return integer
+     */
+    public function getTotalCount()
+    {
+        $this->totalCount = count($this->items);
```
DavidGarciaCat: If I call this method 100 times I will run the `count` function 100 times, AKA performance impact. Due the result won't change, you can just set the `totalCount` value inside the constructor

src/Mailgun/Model/Suppression/Unsubscribe/IndexResponse.php
```diff
@@ -21,11 +21,25 @@
...
-    private $items;
+    private $items = [];
+    
+    /**
+     * Store the total number of Unsubscribe items
+     * 
+     * @see Mailgun/Model/Suppression/Unsubscribe/Unsubscribe
+     * @var integer
+     */
+    private $totalCount = 0;
```
DavidGarciaCat: Please refer to my comment on `getter` method. Thanks,

src/Mailgun/Model/Suppression/Unsubscribe/IndexResponse.php
shakaran: Are you sure that the count will not change? For example if I add items manually or I add items (because maybe I want merge a list of mails to unsubscribe and after add more to resent to the API) ?

src/Mailgun/Model/Suppression/Unsubscribe/IndexResponse.php
shakaran: The advantage of initializing properties outside of the constructor is that someone reading your code will immediatly know its a default value.

Also for automatic analysis or help in IDE.

I don't think there is much of a difference when it comes to performance and it indeed helps to newbies from IDE or so

src/Mailgun/Model/Suppression/Unsubscribe/IndexResponse.php
shakaran: Also, if it doesn't change the value (following your thoughts) it is really more probably call 1000 times to getTotalCount() if it doesnt change the value?

src/Mailgun/Model/Suppression/Unsubscribe/IndexResponse.php
DavidGarciaCat: AFAIK IDE's are using the annotation against the value, so `Unsubscribe[]` defines an array of `Unsubscribe` objects, and a developer will know too that we're dealing with an array.

src/Mailgun/Model/Suppression/Unsubscribe/IndexResponse.php
DavidGarciaCat: IMO given this object is instantiated via other SDK calls when we run an API request (via `public static function create(array $data)` static method) then we can calculate the expected figure when we're creating the object.

In the end, constrictor is a `private` method and this object is mapping data that was returned by the system, so I don't think this is really necessary, and I don't think it will change.

In fact, if you really need to add new / non-covered behaviour, then you can always set-up your own `Decorator` to deal with your specific request, but - again, IMHO - I don't think this update is required here.

shakaran:
@Nyholm please could check for tell me which exactly changes are needed so the PR could be merged?

src/Mailgun/Model/Suppression/Unsubscribe/IndexResponse.php
Nyholm: This class is immutable so nothing will change. 

Im fine with calling count 100 times, the performance impact is minimal compared to the HTTP request to mailgun server that we just did. 

However, it seams weird to have the `$this->totalCount`. That variable is never properly used... 

I wish that the `getTotalCount` functions should be rewritten to "make more sense". Either we just do `return count($this->items);` or we store the value in `$this->totalCount` and never writes to that variable again.

DavidGarciaCat:
Hey there,

@shakaran do you have some time to review last comment and provide an update, before we proceed with a new review?

Cheers,

Nyholm:
I made a small change to this PR and I think it is good now.

PR #394, opened 2017-08-30
Implement getId(), setAddress() and add doc

shakaran:
This should cover part 4 of issue #390 

I add doc in general and implement getId() and setAddress(). Also some useful changes as return $this in set methods that allow chain the setters for example:

```
$unsubscribe = (new Unsubscribe('someemail@example.com'))
                         ->setTag('my_tag')
                         ->setId('000000001')
                         ->setCreationDate(\Datetime('NOW'));
```

### DIFF
--- a/src/Mailgun/Model/Suppression/Unsubscribe/Unsubscribe.php
+++ b/src/Mailgun/Model/Suppression/Unsubscribe/Unsubscribe.php
@@ -15,31 +15,54 @@ namespace Mailgun\Model\Suppression\Unsubscribe;
 class Unsubscribe
 {
     /**
+     * The unsubscribe event identifier
+     * 
      * @var string
      */
-    private $address;
-
+    private $id = NULL;
+    
     /**
+     * The unsubscribe email address
      * @var string
      */
-    private $tag;
+    private $address = NULL;
 
     /**
+     * Tag to unsubscribe from, use * to unsubscribe address from domain
+     * 
+     * Note: If NULL the value means * (all for a domain)
+     * 
+     * @var string
+     */
+    private $tag = NULL;
+
+    /**
+     * The creation date for the event
+     * 
      * @var \DateTime
      */
-    private $createdAt;
+    private $createdAt = NULL;
 
     /**
      * @param string $address
      */
-    private function __construct($address)
+    private function __construct($address = NULL)
     {
         $this->address = $address;
         $this->createdAt = new \DateTime();
     }
 
     /**
-     * @param array $data
+     * Create a Unsubscribe object based in info provided in $data array
+     * 
+     * Fields allowed:
+     * 
+     * - id: The unsubscribe event identifier
+     * - address: The unsubscribe email address
+     * - tag: Tag to unsubscribe (use * or NULL to unsubscribe address from domain)
+     * - created_at: The creation date for the event
+     * 
+     * @param array $data the info provided
      *
      * @return Unsubscribe
      */
@@ -47,9 +70,14 @@ class Unsubscribe
     {
         $unsubscribe = new self($data['address']);
 
+        if (isset($data['id'])) {
+            $unsubscribe->setId($data['id']);
+        }
+        
         if (isset($data['tag'])) {
             $unsubscribe->setTag($data['tag']);
         }
+        
         if (isset($data['created_at'])) {
             $unsubscribe->setCreatedAt(new \DateTime($data['created_at']));
         }
@@ -57,7 +85,31 @@ class Unsubscribe
         return $unsubscribe;
     }
 
+	/**
+	 * Get the event identifier
+	 * 
+     * @return string The event identifier
+     */
+    public function getId()
+    {
+        return $this->id;
+    }
+    
     /**
+     * Set the event identifier 
+     * 
+     * @param string $id The event identifier
+     */
+    private function setId($id = NULL)
+    {
+        $this->id = $id;
+        
+        return $this;
+    }
+    
+    /**
+     * Get the email address unsubscribed
+     * 
      * @return string
      */
     public function getAddress()
@@ -66,6 +118,24 @@ class Unsubscribe
     }
 
     /**
+     * Set the email address unsubscribed
+     * 
+     * @param string $address The email address unsubscribed
+     */
+    private function setAddress($address = NULL)
+    {
+        $this->address = $address;
+        
+        return $this;
+    }
+    
+    /**
+     * Get the tag 
+     * 
+     * Note: * means all unsubscribe address from domain
+     * 
+     * If NULL the value means * (all for a domain)
+     * 
      * @return string
      */
     public function getTag()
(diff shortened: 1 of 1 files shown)


src/Mailgun/Model/Suppression/Unsubscribe/Unsubscribe.php
```diff
@@ -15,49 +15,101 @@
...
     {
         $this->address = $address;
         $this->createdAt = new \DateTime();
     }
 
     /**
-     * @param array $data
+     * Create a Unsubscribe object based in info provided in $data array
+     * 
+     * Fields allowed:
```
Nyholm: Good. Do you know of the PSR5 syntax?

https://github.com/php-http/cache-plugin/blob/master/src/CachePlugin.php#L50-L62

src/Mailgun/Model/Suppression/Unsubscribe/Unsubscribe.php
```diff
@@ -15,49 +15,101 @@
...
+    {
+        return $this->id;
+    }
+    
+    /**
+     * Set the event identifier 
+     * 
+     * @param string $id The event identifier
+     */
+    private function setId($id = NULL)
```
Nyholm: Why should this parameter be optional?

src/Mailgun/Model/Suppression/Unsubscribe/Unsubscribe.php
```diff
@@ -66,6 +118,24 @@ public function getAddress()
     }
 
     /**
+     * Set the email address unsubscribed
+     * 
+     * @param string $address The email address unsubscribed
+     */
+    private function setAddress($address = NULL)
```
Nyholm: Same here, why is this optional?

src/Mailgun/Model/Suppression/Unsubscribe/Unsubscribe.php
```diff
@@ -74,14 +144,24 @@ public function getTag()
...
+     * If NULL the value means * (all for a domain)
+     * 
+     * @param string $tag The tag
      */
-    private function setTag($tag)
+    private function setTag($tag = NULL)
     {
         $this->tag = $tag;
+        
+        return $this;
```
Nyholm: Do we need a fluent api for these?

src/Mailgun/Model/Suppression/Unsubscribe/Unsubscribe.php
shakaran: Oww! Awesome. I will update it. I didn't know about PSR5

src/Mailgun/Model/Suppression/Unsubscribe/Unsubscribe.php
shakaran: It is a way to guarantee at least NULL if not provided or received by the rest client or so. Also since was not previously loaded or used avoid future errors with people code. The really nice thing is if I could put private function setId(string $id = NULL) : Unsubscribe  (using the PHP 7.1 syntax), but it seems that 5.5 and 5.6 are still supported.

src/Mailgun/Model/Suppression/Unsubscribe/Unsubscribe.php
shakaran: Yes, since you could be interested in init empty object or even "clear" the object or attribute after received the data

src/Mailgun/Model/Suppression/Unsubscribe/Unsubscribe.php
shakaran: I think that doesn't hurt, in fact could be good add in other models too. It is the symfony way in setters as far I understand and a lot people expect from moderm apis the chanining of setters ;)

src/Mailgun/Model/Suppression/Unsubscribe/Unsubscribe.php
DavidGarciaCat: IMHO if:

```php
if (isset($data['id'])) {
    $unsubscribe->setId($data['id']);
}
```

Then there's no need to allow this

src/Mailgun/Model/Suppression/Unsubscribe/Unsubscribe.php
DavidGarciaCat: I don't think it should be optional, based on the API docs

src/Mailgun/Model/Suppression/Unsubscribe/Unsubscribe.php
DavidGarciaCat: I don't think we need to return `this` here, but up to you

src/Mailgun/Model/Suppression/Unsubscribe/Unsubscribe.php
```diff
@@ -90,10 +168,14 @@ public function getCreatedAt()
     }
 
     /**
-     * @param \DateTime $createdAt
+     * The creation date for the event
+     * 
+     * @param \DateTime $createdAt The creation date for the event
      */
-    private function setCreatedAt(\DateTime $createdAt)
+    private function setCreatedAt(\DateTime $createdAt = NULL)
```
DavidGarciaCat: I don't think this should be optional, based on existing api docs

src/Mailgun/Model/Suppression/Unsubscribe/Unsubscribe.php
shakaran: @DavidGarciaCat my idea (if the PR is accepted) is make new PR iterating over all the models and classes and add the chaining feature to every set method

src/Mailgun/Model/Suppression/Unsubscribe/Unsubscribe.php
shakaran: The id attribute is on the server spec, but not always is set, so why is not needed?

DavidGarciaCat: What I mean is:
- If we don't provide a value for a given property, ten by default the value is `null`
- If we don't have the `id` key on the array, then there's no need to call the setter
- If the setter is just setting the value then there's no need to have the setter - due we should not change IDs of external resources, in this case some data that comes from an API - so you can set the value in this way: `$unsubscribe->id = $data['id']`

src/Mailgun/Model/Suppression/Unsubscribe/Unsubscribe.php
DavidGarciaCat: @shakaran although I know what you mean, I still don't see the need to add `return $this;` everywhere, specially when we're dealing with `private` methods.

Also, what happens if - a human mistake - we forget a single `return $this;` and then the SDK throws an error due we can't concatenate the next call?

Decouple matters, and may help to see what you're doing with each object - don't see the problem to use:
```php
$myAwesomeModal->setFirstValue('...');
$myAwesomeModal->setSecondValue('...');

$anotherResource->setAnotherValue('...');
```

Nyholm:
Any updates on this PR?

Issue #298, opened 2017-03-01
multiple fopen, which are never closed


cklm:
Hi, in src/Mailgun/Connection/RestClient.php are two `fopen`-functions, which are not closed. Sending over 1.000 mails in serial line leads to the following error:
`[Http\Client\Exception\NetworkException]
error:02001018:system library:fopen:Too many open files`

Is it possible to safely close the two fopens, or are they needed in another place?

Nyholm:
I believe it is safe to close it. Can you send a PR?

Nyholm:
Closing due inactivity.

PR #395, opened 2017-08-30
Close fopen files in post for RestClient

shakaran:
This is a possible fix for issue #298

### DIFF
--- a/src/Mailgun/Connection/RestClient.php
+++ b/src/Mailgun/Connection/RestClient.php
@@ -310,6 +310,10 @@ class RestClient
             $resource = fopen($filePath, 'r');
         }
 
+        if(is_resource($resource)) {
+            fclose($resource);
+        }
+        
         return [
             'name' => $fieldName,
             'contents' => $resource,


src/Mailgun/Connection/RestClient.php
```diff
@@ -310,6 +310,10 @@ protected function prepareFile($fieldName, $filePath)
             $resource = fopen($filePath, 'r');
         }
 
+        if(is_resource($resource)) {
+            fclose($resource);
```
Nyholm: Are you sure you can close it here? The resource is being used after this point.

shakaran: @Nyholm you are right, well the only point then would before to return in [line 179](https://github.com/shakaran/mailgun-php/blob/cb273daa088f03c79d7f35fb90334d9c1b221694/src/Mailgun/Connection/RestClient.php#L179) the problem is that it should force to iterate over all the array $postFiles and pick the field "contents" to close it. So in terms of perfomance, I am not sure if it will degrade it, what do you think?

shakaran:
@Nyholm any update for this? I am waiting your input

src/Mailgun/Connection/RestClient.php
shakaran: @Nyholm any update? several months has passed and I am still waiting for some feedback/input. it seems that the project has little activity or none commiting new code :(

Nyholm:
Hey. Sorry for the delay. 
What makes me hesitant of this PR are a few things: 
1. No tests to prove its correctness
2. The class is deprecated, should we really add a performance fix in a deprecated class?
3. We are not sure if it is a performance fix
4. Are we really sure that `prepareFile` could return a closed resource? I think I wrote part of that code and Im not sure... 

I think I rather spend some hours and effort in fixing everything that need to be done before 3.0 so we can remove this class.

DavidGarciaCat:
Agree with @Nyholm - better if we focus our efforts on v3 instead of work on `deprecated` code

shakaran:
I am currently experiencing this problem with new dev-master, when I attach 5 files to the message and sending to 5000 people. If the current user has a ulimit of 1000 open files for example. I cannot send this, because it will only send to the first 250 people (250 people with 5 files per person = 1000 open files).
I am trying to close all the opened files (resources stream) with the following:

array_map('fclose', (get_resources('stream')));

But it seems to close the php script too. So this is a current issue in dev master. I am agree on no update the deprecated code, but a new issue should be opened and really important solve this issue, because it limits the potential to send email with attachments in batch mode

DavidGarciaCat:
Hi @shakaran 

IMHO we have inaccurate data here - just an example: what are the file size for each one of those 5? Maybe you're reaching the limit that Mailgun allows?

On the other hand, sorry but it was not clear to me: are you sending 5000 messages at once, or are you splitting the recipients in small batches of 250 and then sending them, like running a loop to deal with each request?

I don't think that close the resource with `fclose` is the right solution here, as it was discussed previously, and you just pointed out another issue with this approach, so I think the right answer is to look for another approach to send those files to this large amount of recipients that you just said.

Cheers,

shakaran:
@DavidGarciaCat 

The size is pretty standard, less than 1 MB or so. Exactly are this files:

File 1:
https://iberopistacho.com/bundles/iberopistachomain/pdf/marco-plantacion.pdf

File 2:
https://iberopistacho.com/bundles/iberopistachomain/pdf/polinizacion/polinizacion-macho-hembra.pdf

File 3:
https://iberopistacho.com/bundles/bundles/iberopistachomain/pdf/profesionales-y-empresas-recomendados-iberopistacho.pdf

File 4:
https://iberopistacho.com/bundles/bundles/iberopistachomain/pdf/prestamo-agroinversion-agrobank-iberopistacho-garantia-personal-8-years.pdf

File 5:
https://iberopistacho.com/bundles/bundles/iberopistachomain/pdf/prestamo-agroinversion-agrobank-iberopistacho-garantia-hipotecaria-15-years.pdf

(All the files are public).

I have a script that fetch the 5000 emails from my user database. I just make a loop (foreach in php), and I sent to the 5000 users.

Since I have the problem with the opened files, I have to put a condition in a for, to start each 250 users, and change manually after the script stop.

I get this working now to sent to the 5000 users, if I execute this lines in each iteration (leaving only 11 files opened, because each iteration close the n-1 file.

```php
$streams = get_resources('stream');

                    if(count($streams) > 10)
                    {
                        echo '  Closing Files opened (' . count($streams) . ')' . PHP_EOL;
                        echo '  ' . array_slice($streams, -2, 1)[0] . PHP_EOL;
                        array_map('fclose', array_slice($streams, -2, 1));
                    }
```

But indeed. I think that it is a problem with fopen or some stream not closed in each send() execution.

My method to send the emails are:

```php
$mgClient = Mailgun::create(self::MAILGUN_KEY);


$attachmentsMail = [
       [
            'filename' => 'marco-plantacion.pdf',
            'filePath' => '/home/iberopis/public_html/bundles/iberopistachomain/pdf/marco-plantacion.pdf'
        ],
        [
            'filename' => 'polinizacion-macho-hembra.pdf',
            'filePath' => '/home/iberopis/public_html/bundles/iberopistachomain/pdf/polinizacion/polinizacion-macho-hembra.pdf'
        ], 
        [
            'filename' => 'Estudio_de_Viabilidad_IberoPistacho.pdf',
            'filePath' => '/home/iberopis/public_html/bundles/iberopistachomain/pdf/Estudio_de_Viabilidad_IberoPistacho.pdf'
        ],
        [
            'filename' => 'profesionales-y-empresas-recomendados-iberopistacho.pdf',
            'filePath' => '/home/iberopis/public_html/bundles/iberopistachomain/pdf/profesionales-y-empresas-recomendados-iberopistacho.pdf'
        ],

        [
            'filename' => 'prestamo-agroinversion-agrobank-iberopistacho-garantia-personal-8-years.pdf',
            'filePath' => '/home/iberopis/public_html/bundles/iberopistachomain/pdf/prestamo-agroinversion-agrobank-iberopistacho-garantia-personal-8-years.pdf'
        ],
        [
            'filename' => 'prestamo-agroinversion-agrobank-iberopistacho-garantia-hipotecaria-15-years.pdf',
            'filePath' => '/home/iberopis/public_html/bundles/iberopistachomain/pdf/prestamo-agroinversion-agrobank-iberopistacho-garantia-hipotecaria-15-years.pdf'
        ], 
    ];
    
        try
        {
            # Make the call to the client.
            $result = $mgClient->messages()->send($domain, [
                'v:customvarcampaign' => $this->CAMPAIGN_NAME,
                'o:tracking' => 'yes',
                'o:tracking-clicks' => 'yes',
                'o:tracking-opens' =>  'yes',
                'o:dkim'  => 'yes',
                'o:campaign' => $this->CAMPAIGN_NAME,
                'from'    => 'Someone <someaddess@domain.com>',
                'to'      => trim($name) . ' <' . trim($email) . '>',
                'subject' => $this->SUBJECT_NAME,
                'o:tag'   => [$this->TAG_NAME],
                'attachment' => $attachmentsMail,
                'html'    => 'some html'
            ]);

            if(!empty($result))
            {
                if($result instanceof SendResponse)
                {
                    echo '  Result ➡ ID: ' . $result->getId() . ' Message: ' . $result->getMessage() . PHP_EOL;

                    $streams = get_resources('stream');

                    if(count($streams) > 10)
                    {
                        echo '  Closing Files opened (' . count($streams) . ')' . PHP_EOL;
                        echo '  ' . array_slice($streams, -2, 1)[0] . PHP_EOL;

                        array_map('fclose', array_slice($streams, -2, 1));
                    }
                }
                else
                {
                        var_dump($result);
                }
            }
            else
            {
                var_dump($result);
            }
        }
        catch(\Exception $e)
        {
            echo $email . ' MailgunSend Exception:' . $e->getMessage() . PHP_EOL . $e->getTraceAsString();
        }
```

PR #401, opened 2017-10-04
Bug fix to correct stats/total API call returning nothing

incy247:
Fixed incorrect isset check in stats/total


incy247:
Duplicate of pull/384

PR #403, opened 2017-10-16
Removed unnecessary new lines

alanaasmaa:
It feels a bit cleaner to me like this.

Or would it look better like this =>
```
// After 1,000 recipients, Batch Message will automatically post your message to 
// the messages endpoint. 
```

### DIFF
--- a/src/Mailgun/Messages/README.md
+++ b/src/Mailgun/Messages/README.md
@@ -99,8 +99,7 @@ Here's how to use Batch Message to easily handle batch sending jobs.
 $mg = new Mailgun("key-example");
 $domain = "example.com";
 
-# Next, instantiate a Message Builder object from the SDK, pass in your sending 
-domain.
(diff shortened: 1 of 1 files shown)


alanaasmaa:
@DavidGarciaCat

I understand the point about screen sizes, but right now its syntax error.

How about 

```

/*
* Loooooong comment
*
*/

```

DavidGarciaCat:
@Nyholm @pirogoeth thoughts on this?

Nyholm:
PSR2 specifies that no line should be over 80 chars. 
This is also what Symfony is using for their docs. Keeping 80 chars per line will make the code/docs easier to review. 

Since these changes is in fact a syntax error and the lines are part of code, I think this is fine. The lines are still < 100 chars. If the lines were longer we should probably do multiline comment as @alanaasmaa suggests.

PR #405, opened 2017-10-17
Correction of the "Readable filename" array record key

tfrajt:
Guidelines described in docs/attachment.md in section "From file path" that "filename" array record Id should be used as an ID for readable name of attached file, however in file "/src/Mailgun/Connection/RestClient.php", specifically in method "prepareFile" where processing of attachments is in fact implemented, is "filename" information in this case read out from the array record with id "remoteName".


tfrajt:
:( ... sorry, I haven't modified test classes ... didn't realized it. I'll modify them too.

PR #408, opened 2017-10-31
Added addToRecipient to BatchMessage class

AimzAmmar:
Added addToRecipient to BatchMessage class, this should work as per documentation and should be able to purge recipients greater than 1,000 automatically without throwing an exception if auto send is on.

### DIFF
--- a/src/Mailgun/Messages/BatchMessage.php
+++ b/src/Mailgun/Messages/BatchMessage.php
@@ -103,6 +103,20 @@ class BatchMessage extends MessageBuilder
         }
         $this->batchRecipientAttributes["$address"] = $variables;
     }
+    
+     /**
+     * @param string     $address
+     * @param array|null $variables
+     * @return mixed
+     */
+    public function addToRecipient($address, $variables = null)
+    {
+        $variables = is_array($variables) ? $variables : [];
+
+        $this->addRecipient('to', $address, $variables);
+
+        return end($this->message['to']);
+    }
 
     /**
      * @param array $message


src/Mailgun/Messages/BatchMessage.php
```diff
@@ -103,6 +103,20 @@ protected function addRecipient($headerName, $address, $variables)
         }
         $this->batchRecipientAttributes["$address"] = $variables;
     }
+    
+     /**
+     * @param string     $address
+     * @param array|null $variables
+     * @return mixed
+     */
+    public function addToRecipient($address, $variables = null)
```
DavidGarciaCat: If `$variables` should be an `array` then we should type hint it properly:
`array $variables = null`

src/Mailgun/Messages/BatchMessage.php
```diff
@@ -103,6 +103,20 @@ protected function addRecipient($headerName, $address, $variables)
         }
         $this->batchRecipientAttributes["$address"] = $variables;
     }
+    
+     /**
+     * @param string     $address
+     * @param array|null $variables
+     * @return mixed
```
DavidGarciaCat: We should never `return mixed` due it is confusing and it may result in a mess in a long-term.

Can you please check what's the content that we are storing and then change the type hint here?

Thanks,

DavidGarciaCat: As a minor change, we should have an empty line between `@param` and `@return`:
```php
/**
 * @param string     $address
 * @param array|null $variables
 *
 * @return ...
 */
```

AimzAmmar: I am sorry I am working on an extensive platform and I haven't tested it thoroughly yet, its still under development. You can discard the changes if you feel so but I feel like this is how it should be built otherwise instead of sending queued messages when the recipients reach 1,000 it will throw an exception.

I will update you along with Unit tests once I am done with the project though it will be taking quite some time and this works as a small aspect of the project.

Thanks

Nyholm:
This is added in master.

Issue #417, opened 2017-12-02
Interfaces names should end with "Interface"


DavidGarciaCat:
IMHO any `Interface` file should end with `Interface` for better clarity.

We have currently 3 interfaces that are not ending with it:
- `src/Mailgun/Hydrator/Hydrator.php`
- `src/Mailgun/Model/ApiResponse.php`
- `src/Mailgun/Model/PagingProvider.php`

I would suggest to rename them - both files and code - to include `Interface`, then update all those files that are implementing these interfaces to make sure they are matching with the new names.

Nyholm:
Should we close this?

DavidGarciaCat:
Well... as discussed this is more a Symfony standard than PHP standard, so I guess that we can close it

PR #418, opened 2017-12-04
Mailgun #417 Interfaces names should end with "Interface"

DavidGarciaCat:
As per my own inspection running SensioLabs Insight:
An `Interface` file should end with `Interface` for better clarity.

This PR should fix https://github.com/mailgun/mailgun-php/issues/417

### DIFF
--- a/src/Mailgun/Api/HttpApi.php
+++ b/src/Mailgun/Api/HttpApi.php
@@ -12,7 +12,7 @@ namespace Mailgun\Api;
 use Http\Client\Exception as HttplugException;
 use Http\Client\HttpClient;
 use Mailgun\Exception\UnknownErrorException;
-use Mailgun\Hydrator\Hydrator;
+use Mailgun\Hydrator\HydratorInterface;
 use Mailgun\Hydrator\NoopHydrator;
 use Mailgun\Exception\HttpClientException;
 use Mailgun\Exception\HttpServerException;
@@ -32,7 +32,7 @@ abstract class HttpApi
     private $httpClient;
 
     /**
-     * @var Hydrator
+     * @var HydratorInterface
      */
     protected $hydrator;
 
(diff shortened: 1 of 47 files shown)


Nyholm:
This is symfony coding standards. It is not part of PSR2. 

This is a BC breaking change. If this is merged the next version has to be 3.0.

DavidGarciaCat:
Not a problem! ;-)

Nyholm:
I've reviewed the PR and I think it looks good. 

However, since merging this PR will break BC we should ask ourselves: "Is this PR worth releasing 3.0 over?". I do not think so, but I would love to merge this when the arguments for 3.0 becomes stronger. 

We should create a better roadmap. There are a few things that need to be done. On the top of my head: BatchMessages and documentation.

DavidGarciaCat:
Hey @Nyholm as I said on my last message, "not a problem", so happy to discuss potential changes for the v3 roadmap. By other side, I still need to work on the Email Validator update, hope to do it soon, I think I'll have some time next days

TemirkhanN:
@DavidGarciaCat since it breaks backward compatibility and you still need it badly I suggest to make steps iterably keeping backward compatibility. Kind of

```PHP

interface HydratorInterface extends Hydrator {}

/**
 * @deprecated will be removed in 3.0.0
 */
interface Hydrator {...}
```

DavidGarciaCat:
Hy @TemirkhanN 

Don't worry for the BC, as @Nyholm pointed this PR can be merged as part of the v3 SDK, a new major version means that we might break BC due the removal of deprecated code or the implementation of other changes like - like this one.

Best,

DavidGarciaCat:
As pointed out [here](https://github.com/mailgun/mailgun-php/pull/418#issuecomment-348885125), this is `This is symfony coding standards` so I believe that we can close this PR for now. Let's discuss if we really need this or not.

PR #420, opened 2017-12-04
Clean-up duplication

Eugst:
Deleted `Model/Domain/VerifyResponse` and all references changed to `Model/Domain/CreateResponse`.


Nyholm:
Yeah, Thank you!

PR #424, opened 2017-12-07
Make sure we do not instantiate an abstract class.

Nyholm:
Fixed minor bug introduced in #420. 

FYI @Eugst, We missed this in the review.


Eugst:
yay... Sorry, my bad... I forgot to push :) it was in a second commit... 
thanks for cleanup! strange that tests didn't caught it.

Nyholm:
No worries at all. Both me and David missed it. I'll tag this fix in a new release now. 

Thank you

DavidGarciaCat:
Nice catch @Nyholm - We'll need to pay more atention

Nyholm:
No worries, That is why we have tests. =)

Issue #191, opened 2016-10-14
Create Email Validation API


pirogoeth:
Includes request dispatcher and any request//response models necessary.

DavidGarciaCat:
If you're going to use this issue to validate email addresses, please consider to use this project https://github.com/egulias/EmailValidator instead of build a completely new validator. This one is a well established package and offers several options to validate email addresses.

Nyholm:
There is a API endpoint for that: https://documentation.mailgun.com/api-email-validation.html

(Even though I believe that egulias email validator should be used all the time.. ) =)

DavidGarciaCat:
Yeah, I know about that endpoint. The reason to use the egulias email validator is just to prevent unnecessary API calls (in the end it means a more healthy service on both sides: local code can return an invalid email address what it means we won't need to send an API request and we will save time, and the API won't get an email validation request that we already know that will fail).

shakaran:
I am curious from this issue, the api email validation offered by Mailgun has some improvement or difference about the egulias library? If not I think that should be deprecated and removed for 3.0. Other option is add as dependency and overwrite the api email validation using the library (to still provide the "feature" inside the mailgun PHP-SDK

DavidGarciaCat:
Hey @shakaran 

This is a pending update, and I will work on it as soon as I can.

My idea is to provide a 1st check via egulias library, and if all is fine then send the request to Mailgun API, due if there's a bad formed email address then doesn't make sense to sent the request to the API validator.

I hope to have some free time the upcoming weeks, and build the updated code.

Cheers,

DavidGarciaCat:
@Nyholm or @pirogoeth feel free to assign me the issue if you want

shakaran:
@DavidGarciaCat LGTM as good solution and save queries to the API too (win-win) to both sides

DavidGarciaCat:
Hey @Nyholm 

I have started to work on the `Email Validator` feature, but I would like to get your POV on it.

The [API docs](https://documentation.mailgun.com/en/latest/api-email-validation.html#email-validation) confirms that the Email Validation endpoints have their own Public / Private keys to send the HTTP requests to the API, so we cannot reuse the same API Key we have for the other HTTP calls due it will throw an error due a wrong API Key.

**1st question**

[...]

PR #427, opened 2017-12-08
Mailgun #191 Email Validation

DavidGarciaCat:
Validates an Email Address https://github.com/mailgun/mailgun-php/issues/191

ToDo:
- Confirm with @Nyholm about the best way to proceed with the API keys and implement it.
- Make sure the PHP INTL extension won't be a problem for Spoof Checks https://github.com/mailgun/mailgun-php/issues/191#issuecomment-369764208


Nyholm:
Sorry for the super late reply. 
I know we discussed this offline before, anyhow, I'll write this down just for the record. 

I like this PR. But I dont think the API client should validate emails itself. It should just pass validation calls to the server.

composer.json
```diff
@@ -8,7 +8,8 @@
         "php-http/message": "^1.0",
         "php-http/client-common": "^1.1",
         "php-http/discovery": "^1.0",
-        "webmozart/assert": "^1.2"
+        "webmozart/assert": "^1.2",
+        "egulias/email-validator": "^2.1"
```
Nyholm: This should be removed.

DavidGarciaCat: Yep, agree. I’ll proceed with the steps that we discussed the last time we met and discussed this case. I’ll keep you posted for a final review.

Nyholm:
Fixed the merged conflict. =)

DavidGarciaCat:
Hey @Nyholm 

I've needed to rebase due to a conflict didn't allow to merge; we can merge the PR once the new Travis Built is passing.

I think that more testing is needed before to release v3, especially due to the recent updates (we have updated a lot of files0; I wouldn't like to release v3 and in a couple of days push a bug fix.

Cheers,

Nyholm:
Wohoo!

PR #432, opened 2017-12-19
feat(unsubscribe): Handle tags in Unsubscribe v3 api.

qpautrat:
Hi,

Actually tags from v3 are not fulfilled.
This PR fixes the issue without introducing BC.

### DIFF
--- a/src/Mailgun/Connection/Exceptions/GenericHTTPError.php
+++ b/src/Mailgun/Connection/Exceptions/GenericHTTPError.php
@@ -17,6 +17,7 @@ use Mailgun\Exception;
 class GenericHTTPError extends \Exception implements Exception
 {
     protected $httpResponseCode;
+
     protected $httpResponseBody;
 
     public function __construct($message = null, $response_code = null, $response_body = null, $code = 0, \Exception $previous = null)
--- a/src/Mailgun/Constants/Api.php
+++ b/src/Mailgun/Constants/Api.php
@@ -15,10 +15,16 @@ namespace Mailgun\Constants;
 class Api
 {
     const API_USER = 'api';
+
     const SDK_VERSION = '1.7';
+
     const SDK_USER_AGENT = 'mailgun-sdk-php';
+
     const RECIPIENT_COUNT_LIMIT = 1000;
+
     const CAMPAIGN_ID_LIMIT = 3;
+
     const TAG_LIMIT = 3;
+
     const DEFAULT_TIME_ZONE = 'UTC';
 }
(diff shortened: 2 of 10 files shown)


DavidGarciaCat:
Hey @qpautrat 

Can you please explain why do we need to add a `tags` property (plural)? I don't think this is covered by the API based on the official docs here:

https://documentation.mailgun.com/en/latest/api-suppressions.html#view-a-single-unsubscribe

As you can see on the API docs, the `Suppression` method to get a single `Unsubscribed` user is returning just a single `tag` as String, but not a `tags` array of strings:

```js
{
  "address": "alice@example.com",
  "tag": "*",
  "created_at": "Fri, 21 Oct 2011 11:02:55 GMT"
}
```

Also the `POST` method expects an optional `tag` that must be a string, not an array:

screen shot 2017-12-19 at 11 14 26

Can you please put me an example for this update? I don't see why we need to add the `tags` property here, due is not aligned with the API docs.

Cheers,

qpautrat:
Because documentation is invalid.

If you run following command `curl -s --user 'api:API_KEY' -G https://api.mailgun.net/v3/domain/unsubscribes`, this is an example of what you can have in return:

```
{
  "items": [
    {
      "address": "dummy@mailgun.net",
      "tags": [
        "*", "tag1", "tag2"
      ],
      "created_at": "Mon, 18 Dec 2017 19:58:00 UTC"
    }
  ],
  "paging": {
    ...
  }
}
```

Same story for showing single address.

DavidGarciaCat:
Hi @qpautrat 

If that's the case then:

1. The PR os wrong, because is not removing the old / legacy code, and should be updated to replace the useless code and provide the new one. Can you please review the PR to provide the expected code and remove the legacy lines that are no longer in use?

2. You could send a Support Request to Mailgun (https://mailgun.com/app/support/list) and suggest them to update the API docs due this SDK is based on the official docs and they are not aligned with the API response.

Cheers,

qpautrat:
You have right about first point. I just wanted to not break the Suppression API with v2 for whatever reason, but they should not be able to do it anyway. 

I will open a ticket.

qpautrat:
@DavidGarciaCat It is done.
Do you think you will be able to merge this fix soon ?

DavidGarciaCat:
I don't see any reason to delay it as far as:
- the code is reviewed and approved (I can review it later)
- the tools used to check the code returns green status (currently Style CI returns :-1: so if you can add the suggested fixes then it'll be great!)
- we can validate the update with a couple of checks (although PHPUnit is cool a manual review is always a good practice)

Once all of this is done, we can merge the PR, update the ChangeLog to include latest updates and release a new tag, due this PR should be fixing a bug.

qpautrat:
I disagree about test.
It was not my job to fix style in other files than those I modified.
And test is the same story. I'm agree with you when you said we should have more unit test but it is a part of a new whole PR. 
I test the part I changed and is it more than enough in this context.
Huge PR, especially with refactoring is often a bad idea.

qpautrat:
Hello,

Any updates about this one ?

qpautrat:
Well done, thank you @DavidGarciaCat !

PR #438, opened 2018-01-23
use get_class() instead of ::class

dbergan:
Keeps the code compatible for PHP < 5.5


DavidGarciaCat:
Hi @dbergan 

Do you mind to clarify why should we downgrade the whole SDK to provide support for PHP versions that are lower than PHP 5.5?

Even if we could consider it - that don't think is going to happen due nowadays PHP 7.0 and higher are the ones provided by default on new operating systems - this PR is not doing what you want. There are several references to code where the requirement is PHP 5.5.

Cheers,

DavidGarciaCat:
Closing this PR due the following reasons:
- No activity for more than 1 month
- Incomplete piece of code
- "Useless" update due the minimum PHP version for this SDK is PHP 5.5 and newer versions

PR #455, opened 2018-04-06
413 error proper handle. (This will fix #450)

### LINKED ISSUE #450: Case condition does not work
https://github.com/mailgun/mailgun-php/blob/ef3dc6b8bf78b1f760da16b6860c4072c798d286/src/Mailgun/Api/HttpApi.php#L98

Emails returning with a HTTP Status of 413, which I have worked out corresponds to having a too big total attachment size, were not throwing the exception at this condition. The exception that was being thrown was the UnknownErrorException.

My recommended solution is to throw `throw HttpServerException::serverError($statusCode);` as the default instead. Otherwise all that is required is brackets around the condition.

See: https://stackoverflow.com/questions/7801175/conditional-switch-statements-in-php

aradoje:


aradoje:
I haven't referenced the issue correctly, did the PR based on issue #450 . I will look it up in the docs now.

aradoje:
There is no 413 error code in mailgun docs, just the one already covered in the method.
https://documentation.mailgun.com/en/latest/api-intro.html#errors
Guess I wasted some time :)

Issue #456, opened 2018-04-06
Add Mailing Lists support


Etmolf:
So i'am seeing 
`stats()` 
`domains()` 
`tags()` 
`events()` 
`routes()` 
`webhooks()`

and so on.

What about `lists()` ? so we can easily interact with the Mailing Lists API
Is this alsready on the roadmap or can/should i create a PR for this?

Shouldn't be too hard to implement as the underlying infrastructure for these "Api" classes already exists.

DavidGarciaCat:
Duplicated issue, it was included here https://github.com/mailgun/mailgun-php/issues/442

MACscr:
Im confused. If this is the official SDK, why isnt it keeping up with the API capabilities? Frustrating that there is no mailing list support. =(

DavidGarciaCat:
Hi @MACscr 

This SDK is maintained by collaborators and contributors, and we work on this when we have some time.

You can send a pull request if you want to expedite a new feature, such the new calls for the mailing lists. In the meantime, we didn't remove the old methods so you could use them to make calls and manage your lists; we will remove the old methods once the v3 is released.

Hope this makes sense.

Cheers,

PR #457, opened 2018-04-06
Add Mailing Lists support

Etmolf:
Added most of the functionality for the mailing lists api

Issue: #456 

I could use some help with the tests...

This is my first contribution to github ever
So please do not be angry if something does not fit, i am open for advice

### DIFF
--- a/src/Mailgun/Api/MailingList.php
+++ b/src/Mailgun/Api/MailingList.php
@@ -254,7 +254,15 @@ class MailingList extends HttpApi
     {
         Assert::stringNotEmpty($list);
         Assert::isArray($members);
-        Assert::maxCount($members, 1000);
+
+        // workaround for webmozart/asserts <= 1.2
+        if (count($members) > 1000) {
+            throw new InvalidArgumentException(sprintf('Expected an Array to contain at most %2$d elements. Got: %d',
+                count($members),
+                1000
+            ));
+        }
+
         Assert::oneOf($upsert, ['yes', 'no']);
 
         foreach ($members as $data) {


src/Mailgun/Api/MailingList.php
```diff
@@ -0,0 +1,268 @@
...
+     *
+     * @param int $limit Maximum number of records to return (optional: 100 by default)
+     *
+     * @return PagesResponse|array|ResponseInterface
+     *
+     * @throws \Exception
+     */
+    public function pages($limit = 100)
+    {
+        Assert::integer($limit);
```
DavidGarciaCat: We should validate that I don't give you a negative value (that will be integer anyway, so this rule won't prevent an unexpected behaviour). Can you please add another `Assert` like `Assert::greaterThan($limit, 0);` next to the `Assert::integer($limit);`?

src/Mailgun/Api/MailingList.php
```diff
@@ -0,0 +1,268 @@
...
+
+        $params = compact('limit');
```
DavidGarciaCat: Although it might work, I would say to create the `$params` array as an array, like we have on other API files, so be consistent with the code. Everybody know what is `array()` or `[]` (ideally `[]` due it's the one set by StyleCI) but not everybody might know what is `compact`.

src/Mailgun/Api/MailingList.php
```diff
@@ -0,0 +1,268 @@
...
+     * @param string $access_level List access level, one of: readonly (default), members, everyone
+     *
+     * @return CreateResponse|array|ResponseInterface
+     *
+     * @throws \Exception
+     */
+    public function create($address, $name = null, $description = null, $access_level = 'readonly')
+    {
+        Assert::stringNotEmpty($address);
+        Assert::oneOf($access_level, ['readonly', 'members', 'everyone']);
```
DavidGarciaCat: Missed asserts for `$name` and `$description` arguments

src/Mailgun/Api/MailingList.php
```diff
@@ -0,0 +1,268 @@
...
+     * @param string $address      Address for the new mailing list
+     * @param string $name         Name for the new mailing list (optional)
+     * @param string $description  Description for the new mailing list (optional)
+     * @param string $access_level List access level, one of: readonly (default), members, everyone
+     *
+     * @return CreateResponse|array|ResponseInterface
+     *
+     * @throws \Exception
+     */
+    public function create($address, $name = null, $description = null, $access_level = 'readonly')
```
DavidGarciaCat: PHP variables should not be defined as `snake_case` (`$access_level`), they should be `camelCase` (`$accessLevel`).

src/Mailgun/Api/MailingList.php
DavidGarciaCat: I believe that this was added to re-use variable names, like `$access_level` (comments below) however PHP variable names should be written as `camelCase`

src/Mailgun/Api/MailingList.php
```diff
@@ -0,0 +1,268 @@
...
+    {
+        Assert::stringNotEmpty($address);
+        Assert::oneOf($access_level, ['readonly', 'members', 'everyone']);
+
+        $params = compact('address', 'name', 'description', 'access_level');
```
DavidGarciaCat: To be updated, as described above.

src/Mailgun/Api/MailingList.php
```diff
@@ -0,0 +1,268 @@
...
+     * @return UpdateResponse|array|ResponseInterface
+     *
+     * @throws \Exception
+     */
+    public function update($address, $parameters = [])
+    {
+        Assert::stringNotEmpty($address);
+        Assert::isArray($parameters);
+
+        foreach ($parameters as $field => $value) {
```
DavidGarciaCat: Incomplete validation. We know the expected keys for `$parameters` because they are listed on the Mailgun API documentation. We should validate them, to make sure we don't provide wrong key names or wrong values (like subarrays, objects or callables).

src/Mailgun/Api/MailingList.php
DavidGarciaCat: I can see the same issue on other actions, can you please update them too?

src/Mailgun/Api/MailingList.php
```diff
@@ -0,0 +1,268 @@
...
+        Assert::stringNotEmpty($list);
+        Assert::stringNotEmpty($address);
+
+        $response = $this->httpGet(sprintf('/v3/lists/%s/members/%s', $list, $address));
+
+        return $this->hydrateResponse($response, ShowMemberResponse::class);
+    }
+
+    /**
+     * Creates (or updates) a member of the mailing list.
```
DavidGarciaCat: `createMember` through `POST` HTTP Method is just to create - not update - records. Updates are managed through the `PUT` HTTP Method and the `updateMember` method.

src/Mailgun/Api/MailingList.php
Etmolf: I will do it

src/Mailgun/Api/MailingList.php
Etmolf: i extended the switch to include all possible parameters.
unknown `keys` will be catched in the `default` case and simply removed.

src/Mailgun/Api/MailingList.php
```diff
@@ -0,0 +1,307 @@
...
+
+        return $this->hydrateResponse($response, DeleteResponse::class);
+    }
+
+    /**
+     * Returns a paginated list of members of the mailing list.
+     *
+     * @param string $address    Address of the mailing list
+     * @param int    $limit      Maximum number of records to return (optional: 100 by default)
+     * @param mixed  $subscribed `yes` to lists subscribed, `no` for unsubscribed. list all if null
```
DavidGarciaCat: `mixed` type should be updated to match with the expected type on the PHP Docs. In this case, we could set `string|null` due the fact that we expect a string or a non-defined value.

src/Mailgun/Api/MailingList.php
```diff
@@ -0,0 +1,307 @@
...
+    public function createMember($list, $address, $name = null, $vars = [], $subscribed = 'yes', $upsert = 'no')
+    {
+        Assert::stringNotEmpty($list);
+        Assert::stringNotEmpty($address);
+        Assert::nullOrStringNotEmpty($name);
+        Assert::isArray($vars);
+        Assert::oneOf($subscribed, ['yes', 'no']);
+        Assert::oneOf($upsert, ['yes', 'no']);
+
+        $vars = json_encode($vars);
```
DavidGarciaCat: Although it works, personally I don't like to reuse the same variable to change multiple types. At the beginning, this variable is an `array`, and then it's converted to a JSON `string` (assuming that the `json_encode` function has worked).

I would prefer to use a 2nd variable to store the encoded content (like `$jsonVars` to keep it simple), and then [add a checker](https://secure.php.net/manual/en/function.json-last-error.php) to confirm the last JSON method didn't throw an error. If an error was returned then we should throw our Exception to alert the developer.

DavidGarciaCat: On a 2nd thought I am not sure if this is useless, because the Payload sent to the API should be converting all parameters to a single JSON string, including the arguments. I think this line doesn't make sense at the current stage.

src/Mailgun/Api/MailingList.php
```diff
@@ -0,0 +1,307 @@
...
+        foreach ($parameters as $field => $value) {
+            switch ($field) {
+                case 'address':
+                case 'name':
+                    Assert::stringNotEmpty($value);
+
+                    break;
+                case 'vars':
+                    Assert::isArray($value);
+                    $parameters['vars'] = json_encode($value);
```
DavidGarciaCat: Same case than the one described above.

src/Mailgun/Api/MailingList.php
```diff
@@ -0,0 +1,307 @@
...
+
+                    break;
+                case 'subscribed':
+                    Assert::oneOf($value, ['yes', 'no']);
+
+                    break;
+                default:
+                    unset($parameters[$field]);
```
DavidGarciaCat: Same case than the one described above.

src/Mailgun/Model/MailingList/CreateMemberResponse.php
```diff
@@ -0,0 +1,73 @@
...
+        return new self($member, $message);
+    }
+
+    /**
+     * CreateMemberResponse constructor.
+     *
+     * @param Member $member
+     * @param string $message
+     */
+    public function __construct(Member $member, $message)
```
DavidGarciaCat: Responses should be providing a `private constructor` due the response is always created through the `create named constructor`

src/Mailgun/Model/MailingList/CreateMemberResponse.php
```diff
@@ -0,0 +1,73 @@
...
+     */
+    private $message;
+
+    public static function create(array $data)
+    {
+        $member = null;
+
+        if (isset($data['member'])) {
+            $member = Member::create($data['member']);
+        }
```
DavidGarciaCat: Although this is not a blocker, to be consistent with the rest of the API, I would suggest to define the `$member` variable in a single row:

```php
$member = isset($data['member']) ? Member::create($data['member']) : null;
```

Same case for any other parameter handled by Model and Response files, so we will keep the same logic and format everywhere.

src/Mailgun/Model/MailingList/DeleteMemberResponse.php
```diff
@@ -0,0 +1,73 @@
...
+        return new self($member, $message);
+    }
+
+    /**
+     * DeleteMemberResponse constructor.
+     *
+     * @param Member $member
+     * @param string $message
+     */
+    public function __construct(Member $member, $message)
```
DavidGarciaCat: This should be a private constructor, as described above

src/Mailgun/Model/MailingList/DeleteResponse.php
```diff
@@ -0,0 +1,73 @@
...
+        return new self($address, $message);
+    }
+
+    /**
+     * DeleteResponse constructor.
+     *
+     * @param string $address
+     * @param string $message
+     */
+    public function __construct($address, $message)
```
DavidGarciaCat: This should be a private constructor too

src/Mailgun/Model/MailingList/MailingList.php
```diff
@@ -0,0 +1,131 @@
...
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+namespace Mailgun\Model\MailingList;
+
+/**
+ * @author Michael Münch <helmchen@sounds-like.me>
+ */
+final class MailingList
```
DavidGarciaCat: Due this modal is part of the `Response` and given the fact that we need to implement the `create` static method, is not a bad idea if you add the `implements ApiResponse` sentence here. In this way we will have an uniform format for all files that are returned as part of any request.

Same case for other models should be implemented too please. Thanks!

src/Mailgun/Model/MailingList/MailingList.php
```diff
@@ -0,0 +1,131 @@
...
+{
+    /**
+     * @var string
+     */
+    protected $name;
```
DavidGarciaCat: There is no need to set these properties as `protected`, this class is flagged as `final` so all of them can be converted to `private`.

We have this case on other models too. If you can update them at the same time please would be great! Thanks,

src/Mailgun/Model/MailingList/ShowResponse.php
```diff
@@ -0,0 +1,47 @@
...
+        $list = null;
+
+        if (isset($data['list'])) {
+            $list = MailingList::create($data['list']);
+        }
+
+        return new self($list);
+    }
+
+    public function __construct(MailingList $list)
```
DavidGarciaCat: This should be a private constructor

src/Mailgun/Model/MailingList/UpdateMemberResponse.php
```diff
@@ -0,0 +1,73 @@
...
+        return new self($member, $message);
+    }
+
+    /**
+     * UpdateMemberResponse constructor.
+     *
+     * @param Member $member
+     * @param string $message
+     */
+    public function __construct(Member $member, $message)
```
DavidGarciaCat: This should be a private constructor

src/Mailgun/Api/MailingList.php
Etmolf: > but - in the end - the record is not added because it already exists.

The record actually get's updated in the process and the Response is a valid CreateMemberResponse with all the details about the "added" Member. I am well aware of the REST principles, but iam not sure how to handle this any other way - unless we drop the `$upsert` Parameter and don't allow this `feature` of the Mailgun API. If we do allow it, then we also have to mention it in the doc blocks, right? :) In the end, Mailgun violates the principles, not us :P

src/Mailgun/Api/MailingList.php
Etmolf: that sounds good to me - I will think about what exception you take the best, in doubt probably InvalidArgument?

src/Mailgun/Api/MailingList.php
```diff
@@ -0,0 +1,307 @@
...
+ * @author Michael Münch <helmchen@sounds-like.me>
+ */
+class MailingList extends HttpApi
+{
+    /**
+     * Returns a paginated list of mailing lists on the domain.
+     *
+     * @param int $limit Maximum number of records to return (optional: 100 by default)
+     *
+     * @return PagesResponse|array|ResponseInterface
```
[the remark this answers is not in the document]
Etmolf: What about if you use the ArrayHydrator? :)

src/Mailgun/Model/MailingList/MailingList.php
Etmolf: You're right. But maybe an ApiModel interface would be more useful here in the future? Because, strictly speaking, it's actually a model, not a response itself, but that's just conceptual

DavidGarciaCat: That makes sense to me, however I believe it's not a good idea to add a new `ApiModel` to define the same static method, so I believe that a new interface extending a previous one would be better, so in the end both `ApiResponse` and `ApiModel` might extend from a parent interface that might define the static method.

src/Mailgun/Api/MailingList.php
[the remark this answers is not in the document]
DavidGarciaCat: I don't think it might be the case based on the current code, our Hydrator returns a `ResponseInterface` that it's implemented by our models so in the end we will get an instance of the object that we're passing as an argument of the Hydrator.

src/Mailgun/Api/MailingList.php
DavidGarciaCat: Maybe this one, as a generic answer: https://github.com/mailgun/mailgun-php/blob/master/src/Mailgun/Exception/InvalidArgumentException.php

src/Mailgun/Api/MailingList.php
DavidGarciaCat: 🤔 

@Nyholm what do you think? In fact, both points are good...

src/Mailgun/Api/MailingList.php
Etmolf: perfect! that's what i'am already using in the latest commit, so it should fit.

src/Mailgun/Api/MailingList.php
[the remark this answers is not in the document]
Etmolf:
```
final class ArrayHydrator implements Hydrator
{
    /**
     * @param ResponseInterface $response
     * @param string            $class
     *
     * @return array // <==
     */
```

I can adjust the @return statements that is not a problem, I was just not sure because I had seen that in the other classes as well.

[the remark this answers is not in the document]
DavidGarciaCat: 🤔 Nice point, I'll take a look to confirm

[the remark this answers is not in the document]
DavidGarciaCat: I think it won't be a problem because, for this case, we use the Model Hydrator, meaning that the provided class is the one that sets the output:

https://github.com/mailgun/mailgun-php/blob/master/src/Mailgun/Hydrator/ModelHydrator.php#L42-L46

Etmolf:
I'll see that I make a few tests, but could certainly use help. TDD is one of the things that have been on my list for years, but never had priority due to time constraints. Now it looks a bit different and I will take a closer look at it.

src/Mailgun/Api/MailingList.php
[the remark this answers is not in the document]
Etmolf: the default Hydrator is the ModelHydrator yes .. but we can override the Hydrator during the construction of the Mailgun Object and if so, the return type will vary between a Response implementing ResponseInterface and a plain array in case of the ArrayHydrator. Or did i miss something here? 

I do not want to be picky right now, just to understand the whole thing :)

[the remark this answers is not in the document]
DavidGarciaCat: Hey @Etmolf 

I am not sure to follow you, because we're expecting to create the Mailgun object through the `create` static method (as described on [Usage](https://github.com/mailgun/mailgun-php/blob/master/README.md#usage)'s section on README file), that is expecting just 1 argument: the API Key

https://github.com/mailgun/mailgun-php/blob/master/src/Mailgun/Mailgun.php#L105-L115

Original Mailgun's constructor (that was allowing you to customise the Hydrator)  is no longer recommended to instantiate the Mailgun's client, and 2 of the expected arguments have been already deprecated

https://github.com/mailgun/mailgun-php/blob/master/src/Mailgun/Mailgun.php#L64-L86

Based on this, I believe that we can work thinking about the `ModelHydrator` with no other conditionals in mind, that - in fact - is set automatically on the existing public constructor

https://github.com/mailgun/mailgun-php/blob/master/src/Mailgun/Mailgun.php#L85

Hope this makes sense to you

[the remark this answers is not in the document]
Etmolf: Alright :) If you put it that way, that makes sense.

I change the annotations accordingly

src/Mailgun/Api/MailingList.php
```diff
@@ -254,7 +254,15 @@ public function addMembers($list, array $members, $upsert = 'no')
     {
         Assert::stringNotEmpty($list);
         Assert::isArray($members);
-        Assert::maxCount($members, 1000);
```
DavidGarciaCat: Maybe you can try with this: `Assert::allMaxCount($values, $max, $message = '')`

https://github.com/webmozart/assert/blob/master/src/Assert.php#L159

Etmolf: i think 6 lines of code don't hurt until the minimum required version of webmozart/assert get's uped to 1.3 or something like that. afaik it's the only method which requires such validation on the max array length

Etmolf:
I believe I've made all the requested changes, apart from the Create method which also updates. Tests are also available. 

I skipped Integration Tests for now (instead i tested everything on my mailgun account and it works) cause there isn't any way to automatically clean up after the tests, am i right? as we have to use the same methods to clean up which also get tested - pretty tricky in my opinion.

any advices for that? :)

DavidGarciaCat:
Hey @Etmolf 

Sorry for the delay, I've been busy these days.

I will proceed with a new review ASAP and I'll let you know.

Cheers,

Etmolf:
I would like to distance myself personally from Mailgun and will not support anything in this regard. Thanks anyway for your time. Has nothing to do with you.

PR #460, opened 2018-04-09
Cast campaign ID-s to string

### LINKED ISSUE #421: Sending email with integer campaign ID causes exception
```
InvalidArgumentException: "Invalid resource type: integer" in
/files/vendor/guzzlehttp/psr7/src/functions.php:113  
```

I'm using mailgun version 2.3.4.

`MessageBuilder::addCampaignId` supports integer campain ids, according to phpdoc, but the ids are not being cast to string, but integers are not supported when converting to resources in stream_for.

aradoje:
Cast campaign ID-s to string. Test adding of campaign ID-s.

This will fix #421

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The change log describes what is "Added", "Removed", "Changed" or "Fixed" between each release. 
 
+## 2.5.0
+
+### Added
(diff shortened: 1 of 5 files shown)


aradoje:
Thanks

PR #463, opened 2018-05-08
fix: Validate file name length

mrbarletta:
$message could contain a large email text that if sent to is_file it could break with this error:
` is_file(): File name is longer than the maximum allowed path `
This validation prevents using is_file if the $message is longer than the allowed path.

### DIFF
--- a/src/Mailgun/Api/Message.php
+++ b/src/Mailgun/Api/Message.php
@@ -82,7 +82,7 @@ class Message extends HttpApi
         $params['to'] = $recipients;
         $postDataMultipart = $this->prepareMultipartParameters($params);
 
-        if (is_file($message)) {
+         if (strlen($message) < PHP_MAXPATHLEN && is_file($message)) {
             $fileData = ['filePath' => $message];
         } else {
             $fileData = [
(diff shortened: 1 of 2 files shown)


z38:
When the message content is passed to `sendMime()` directly, the maximum length is exceeded very quickly. I think it makes sense to include the check otherwise the option to pass the content directly is only of very limited use.

mrbarletta:
Thanks @DavidGarciaCat  for your prompt reply.

A try-cath will not work due to the SDK using the same function signature to receive a $message variable that could either be a FILE location or a Mime Message, which differ quite a lot in the possible length. 

I fixed the extra space, and I am sorry I didn't read the contributors guidelines, it was late and didn't want to miss the opportunity to raise the issue with a possible fix.

Regarding testing, well I am 100% noob in that area :-1: , but I can give it a shot, may you tell me which test I could copy and start with?

Nyholm:
Thank you. I think this looks good. But we do need at test before we can merge this.

Nyholm:
I added a small test. Thank you

PR #466, opened 2018-06-28
Change exception message of HttpClientException when badRequest to us…

### LINKED ISSUE #467: Response Details Lost
When using the "mailgun-php" the underlying problem such as "Domain already exists in registry" is lost and replaced with a less useful and more generic "The parameters passed to the API were invalid.". This makes it difficult for my integrated service to respond to the end-user in a meaningful way.

Can the underlying message be passed along up the chain to the HttpClientException as the exception message?

Please see this PR for a possible solution:
https://github.com/mailgun/mailgun-php/pull/466

paulrrogers:
…e server content's message to aid resolving problems with user input.

This will fix #467

### DIFF
--- a/src/Mailgun/Exception/HttpClientException.php
+++ b/src/Mailgun/Exception/HttpClientException.php
@@ -57,7 +57,7 @@ final class HttpClientException extends \RuntimeException implements Exception
     {
         // Pass along message from server to aid resolving input problems.
         $content = json_decode($response->getBody()->getContents());
-        $message = ! empty($content['message'])
+        $message = !empty($content['message'])
             ? $content['message']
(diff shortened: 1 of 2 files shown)


ericnkatz:
@DavidGarciaCat I think doing checks on input values prior to sending requests makes a ton of sense but this PR is in regards to bad messaging when we get a 400 response from Mailgun's API. It could be invalid inputs but it could also be a response message like this:

```
{
  message: "Domain already exists in registry. Domain Name: example.com."
}
```

Nyholm:
Thank you for this PR. 
I need a test before I can merge this. 
I also need a test for when the server returns non-json content.

Nyholm:
I've done some minor modifications to this PR. 

Thank you

Issue #471, opened 2018-07-27
New EU region for Mailgun


clicars:
Mailgun has introduced a new EU region. There are two problems:

a) In Mailgun platform when you edit your domain, they notice you that api url is https://api.mailgun.net which is wrong. Correct api endpoint is https://api.eu.mailgun.net
b) Same problem in mailgun-php. By default you use in Mailgun::create method US api endpoint.

At this time for solving this problem I have had to use ::configure instead of ::create in this way:

Mailgun::configure((new HttpClientConfigurator())->setApiKey($key)->setEndpoint('https://api.eu.mailgun.net'));

It is cleaner to use create method. It would be cool if you introduce a second param in create static method to choose your region.

DavidGarciaCat:
Hi @clicars 

The new EU service was announced yesterday, and it requires some adjustments to align the SDK with the new available options.

We can take a look as soon as we can, but if you're feeling comfortable, a Pull Request is always welcomed.

Best,

ryryan:
Hi,
I stumbled across this issue. @clicars' code didn't work out of the box. So in case anyone has this issue whilst we wait for a full EU support, here's mine.
```
$configurator = new HttpClientConfigurator();
$configurator->setEndpoint('https://api.eu.mailgun.net/v3/'.$domain.'/messages');
$configurator->setDebug(true);
$configurator->setApiKey($key);
$mailgun = Mailgun::configure($configurator);
```

pinonirvana:
I tried to use this solution.
it works for sending messages, but not for other tasks.
For instance if I want to retrieve the list of my european domains, the code only retrieves the USA ones.

The code I tried is this:
```
$mgClient = Mailgun::configure((new HttpClientConfigurator())->setApiKey($this->api)->setEndpoint('https://api.eu.mailgun.net'));

 $result = $mgClient->get("domains");      
 print_r($result);
```

DavidGarciaCat:
Hi @pinonirvana 

Can you please try with this other call?
```php
$mgClient = Mailgun::create($apiKey, $endpoint = 'https://api.eu.mailgun.net');

$result = $mgClient->get("domains");      
print_r($result);
```

Cheers,

martijn103:
Hi @DavidGarciaCat, I'm having the same problem. I'm afraid the code you suggested still returns the US domains.

DavidGarciaCat:
Hey @martijn103 

[...]

PR #477, opened 2018-08-04
Add endpoint to the create function

Nyholm:
This will make it easier to choose the new EU region. 

This is related to #471


gavoc:
Could this change be released as a version? EU region is difficult to work with currently.

PR #478, opened 2018-08-05
Add more tests for our APIs

Nyholm:
Sorry for another massive PR. 
This just adds tests to our API endpoints. It also adds some features to the `Api\TestCase` that makes it easier to write tests.


tests/Api/TestCase.php
```diff
@@ -9,7 +9,11 @@
 
 namespace Mailgun\Tests\Api;
 
+use GuzzleHttp\Psr7\Request;
```
Nyholm: This class is extra interesting to review.

PR #485, opened 2018-09-03
MessageBuilder > addBccRecipient > Make $variables optional

ruudk:
This way, it's the same as `addToRecipient` and `addCcRecipient`

### DIFF
--- a/src/Mailgun/Message/MessageBuilder.php
+++ b/src/Mailgun/Message/MessageBuilder.php
@@ -190,7 +190,7 @@ class MessageBuilder
      *
      * @throws TooManyRecipients
      */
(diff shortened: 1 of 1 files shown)


DavidGarciaCat:
Thanks for this PR, but unfortunately this builder is part of a deprecated code that will be removed when releasing v3 so seems useless to add this update. Cheers,

Nyholm:
Does this change also apply to the new improved builder? If so, can you please ad that?

I see no harm in adding this small change, even if we remove the code in next major release.

Nyholm:
Oh, I too got fooled. 
Old namespace: `Mailgun\Messages`
New Namespace: `Mailgun\Message`

Thank you

PR #501, opened 2018-12-13
Fix type error when creating tags 

### LINKED ISSUE #480: DateTime Error in Model/Tag/Tag.php
Uncaught TypeError: Argument 3 passed to Mailgun\Model\Tag\Tag::__construct() must be an instance of DateTime

On Line 40, 

`public function __construct($tag, $description, \DateTime $firstSeen, \DateTime $lastSeen)`

When calling 

`$mailgun->tags()->index($api_url);`

$api_url is my mailgun domain (set correctly, it works with other calls) 

I was unable to get a response for tags until I removed both \DateTime and it fixed all calls to tags().

This may not be the best fix, but it solved the issue and did not throw any other errors.

I am running PHP 7.1.

Arliee:
I just ran into this bug when attempting to call `$mailgun->tags()->index('my.domain.com')`. Fixes #480.

### DIFF
--- a/src/Mailgun/Model/Tag/Tag.php
+++ b/src/Mailgun/Model/Tag/Tag.php
@@ -52,7 +52,7 @@ class Tag
      */
     public static function create(array $data)
     {
-        return new self($data['tag'], $data['description'], $data['first-seen'], $data['last-seen']);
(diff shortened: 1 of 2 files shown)


Arliee:
Apologies for not including this initially.   Hopefully this will suffice?

Arliee:
I've added getters for the two datetimes and added their checks to the test to confirm they are set correctly.

PR #504, opened 2019-01-02
Fluent interface for MessageBuilder and BatchMessage

finwe:
Makes using MessageBuilder and BatchMessage easier.

No potential BC breaks as all changed methods were originally void.

In addition, this fixes a few wrongly specified `@return string` PHPDoc annotations above void methods (by replacing them with now correct `@return MessageBuilder`).

Exceptions are methods `setDeliveryTime` and `addCustomParameter` which return formatted time and custom parameter value. Their return values were not changed and could be changed for a getter in the next major version (BC break).

### DIFF
--- a/src/Mailgun/Message/BatchMessage.php
+++ b/src/Mailgun/Message/BatchMessage.php
@@ -72,6 +72,8 @@ class BatchMessage extends MessageBuilder
      *
      * @throws MissingRequiredParameter
      * @throws TooManyRecipients
+     *
+     * @return BatchMessage
      */
     protected function addRecipient($headerName, $address, array $variables)
     {
@@ -91,6 +93,8 @@ class BatchMessage extends MessageBuilder
         }
 
         $this->batchRecipientAttributes[(string) $address] = $variables;
+
+        return $this;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Nyholm:
We all are confused =)

The deprecated files are in namespace `Mailgun\Messages`.  The new files are in `Mailgun\Message`. 

I am generally against fluent interfaces. But this is actually one of the few places they make sense. What do you say David?

DavidGarciaCat:
Hi both!

I guess that the paths and namespaces have confused me 😓

I'll take another look now

Nyholm:
Hm. How about `MessageBuilder::setDeliveryTime` and `MessageBuilder::addCustomParameter`. They are both public function that returns a string instead of `$this`. 
We should write a `@deprecated` comment about this. In next major we should only return $this, right?

Nyholm:
Oops =)

@finwe could you follow up my comment with another PR? Just to add the comment to the doc block.

DavidGarciaCat:
Due to consistency reasons, makes sense to me to return `$this` for every setter; if we expect to get the object then the getters are probably the best approach

finwe:
Will do.

PR #511, opened 2019-01-05
Added more tests for models

Nyholm:


src/Mailgun/Model/Tag/Tag.php
```diff
@@ -37,7 +37,7 @@ class Tag
      * @param \DateTime $firstSeen
      * @param \DateTime $lastSeen
      */
-    public function __construct($tag, $description, \DateTime $firstSeen, \DateTime $lastSeen)
+    public function __construct($tag, $description, \DateTime $firstSeen = null, \DateTime $lastSeen = null)
```
DavidGarciaCat: Hey @Nyholm 

I can't see any PHP Unit test to create new tags without the dates.

Are we sure this case might happen?

tests/Model/Tag/IndexResponseTest.php
```diff
@@ -0,0 +1,51 @@
...
+<<<'JSON'
+{
+  "items": [
+      {
+        "tag": "red",
+        "description": "red signup button"
+      },
+      {
+        "tag": "green",
+        "description": "green signup button"
```
Nyholm: Here are the test

Nyholm:
Thank you for the review.

PR #515, opened 2019-01-05
Added IP API class, models and test

### LINKED ISSUE #458: Create IP Api
If I look at the current roadmap for v3, this part of the Api seems to be missing too.

There are probably not many use cases where you need it, but to be feature complete at least.
I could do that as soon as the mailing lists are done.

I would then create another pull request if no one has any objections :)

Nyholm:
This will fix #458


Nyholm:
Thank you for the review.

PR #520, opened 2019-01-06
Make sure we dont inherit final classes.

Nyholm:


Nyholm:
WIP

Nyholm:
Ready for review.

DavidGarciaCat:
Not final or abstract - are we sure about this?

Nyholm:
We use those classes. That is why :/

DavidGarciaCat:
I was thinking about being consistent, that's all

Nyholm:
> I was thinking about being consistent, that's all

True. We could remote `ShowResponse` and just let the API return just return this model instead. 
OR, we could do some weird things with traits... This is actually the most simple solution, but im not sure it is the best. 

Lets revisit before tagging 3.0. I will merge now because the build is currently broken.

PR #527, opened 2019-01-06
Added proper PHP CS fixer rules

### LINKED ISSUE #256: Running style lint
It would be helpful to include an example of how to run the style lint for the project so I can be sure my contributions will pass styleci before commiting/pushing. If it isn't already a one-liner, it would be helpful to add a script to the composer.json to run it easily.

Nyholm:
This will fix #256

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,13 +1,29 @@
 <?php
 
-/*
- * In order to make it work, fabpot/php-cs-fixer and sllh/php-cs-fixer-styleci-bridge must be installed globally
- * with composer.
- *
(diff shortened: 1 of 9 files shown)


DavidGarciaCat:
Are you using a global install of CS Fixer? What if my global install is for CS fixer 1.x and the settings here are for 2.x?

Nyholm:
Oh. Btw. You don’t -need- to run it locally. This just gives you the option to do it.

Nyholm:
Thank you for review.

PR #531, opened 2019-01-07
Added test for BC breaks

Nyholm:
This test will fail until we release 3.0 because we currently have made many BC breaks.


DavidGarciaCat:
I'f you don't mind, given the tests are failing as per your announcement would like to keep this PR open until is time to merge, so we can analyse the final code and results before merging

Nyholm:
Yes, I dont think we should merge failing tests. 
I have just disabled this test.. Ie, it will always pass.

Nyholm:
Running `./vendor/bin/roave-backward-compatibility-check || true` did not work when you put that value in a variable. 

I've updated the PR.

PR #535, opened 2019-01-09
Added new CS rules

Nyholm:
Declare strict and no superfluous phpdoc

This PR is generated by php-cs-fixer

### DIFF
--- a/src/Message/BatchMessage.php
+++ b/src/Message/BatchMessage.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /*
  * Copyright (C) 2013 Mailgun
  *
@@ -55,9 +57,7 @@ class BatchMessage extends MessageBuilder
     }
 
     /**
-     * @param string $headerName
-     * @param string $address
-     * @param array  $variables  {
+     * @param array $variables {
      *
      *     @var string $id
      *     @var string $full_name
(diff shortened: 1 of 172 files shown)


tests/Message/BatchMessageTest.php
```diff
@@ -109,15 +111,17 @@ public function testMaximumBatchSize()
...
-    public function testAttributeResetOnEndBatchMessage()
+    public function testRecipientAttributeResetOnEndBatchMessage()
     {
         $this->batchMessage->addToRecipient('test-user@samples.mailgun.org', ['first' => 'Test', 'last' => 'User']);
         $this->batchMessage->setFromAddress('samples@mailgun.org', ['first' => 'Test', 'last' => 'User']);
         $this->batchMessage->setSubject('This is the subject of the message!');
         $this->batchMessage->setTextBody('This is the text body of the message!');
         $this->batchMessage->finalize();
         $message = NSA::getProperty($this->batchMessage, 'message');
-        $this->assertTrue(true, empty($message));
```
Nyholm: This test is poorly written. I also think that this test is wrong. We should never reset the message. That would be catastrophic for the "auto send" feature

src/Message/BatchMessage.php
```diff
@@ -100,26 +100,34 @@ public function finalize(): void
 
         if (empty($this->domain)) {
             throw new RuntimeException('You must call BatchMessage::setDomain before sending messages.');
-        } elseif (empty($message['from'])) {
```
Nyholm: I removed all these elseif

tests/Message/BatchMessageTest.php
DavidGarciaCat: Agree, this test seems wrong

src/Message/BatchMessage.php
DavidGarciaCat: I'm assuming the reason is that there are other checks are in place before to set the values; however, what happens if we don't call those methods? Do you want to rely on the API instead of throwing an Exception and warn the developer?

Nyholm: Oh, We still to all the checks. I just removed `elseif` and replaced them with `if`. It would be the same since the body of the `if` throws an exception.

Nyholm:
> I remember that some time ago we needed to remove then due to unexpected errors. Is it safe to add them now again?

This is PHP7 only. I guess we removed them before because we supported PHP5. 

> Is there a reason to keep these changes that don't look consistent across all files?

I understand it seams inconsistent. The rule is: "If the docs does not bring any value, then it should be removed". 

Ie, This PHP docs does not bring any value, it should be removed: 

```php
/**
 * @return string
 */
function foo(): string
{
  return "foo";
}
```

-----

This docs do bring some value

```php
/**
 * @return User[]
 */
function foo(): array
{
  // ...
}
```

-----

This will also bring some value:

```php
/**
 * @param string $id this is the id of the user
 */
function foo(string $id): void
{
  // ...
}
```

PR #539, opened 2019-01-12
Moved `Mailgun\Model\Route\Response` to `Mailgun\Model\Route`

Nyholm:
To be similar to all the other models.


Nyholm:
The CS failure is not related.

PR #540, opened 2019-01-12
Make sure we dont use any risky tests

Nyholm:


tests/Api/TestCase.php
```diff
@@ -89,14 +89,23 @@ protected function getApiInstance($apiKey = null)
...
-                $this->callback([$this, 'validateRequestUri']),
-                $this->callback([$this, 'validateRequestHeaders']),
-                $this->callback([$this, 'validateRequestBody'])
-            )
-            ->willReturn(new Request('GET', '/'));
+
+        if (null !== $this->requestMethod
+            || null !== $this->requestUri
+            || !empty($this->requestHeaders)
+            || !empty($this->requestBody)
```
DavidGarciaCat: Does really make sense to check non-empty body for a get request?

tests/Api/TestCase.php
```diff
@@ -89,14 +89,23 @@ protected function getApiInstance($apiKey = null)
...
+        ) {
+            $requestClient
+                ->expects($this->once())
```
Nyholm: The idea with this PR is to add this line.

Nyholm:
Thank you

Issue #532, opened 2019-01-07
Add PHP types annotations on models


Nyholm:
All classes in `Mailgun\Model`

PR #541, opened 2019-01-13
Updating models to PHP 7 for Domain and Attachments

Nyholm:
It feels good to remove code. But this process is slow..

I've done Attachments and Domain. Will continue with other models in a separate PR. 

This is part of #532

### DIFF
--- a/src/Model/Domain/AbstractDomainResponse.php
+++ b/src/Model/Domain/AbstractDomainResponse.php
@@ -18,42 +18,19 @@ use Mailgun\Model\ApiResponse;
  */
 abstract class AbstractDomainResponse implements ApiResponse
 {
-    /**
-     * @var string
-     */
     private $message;
-
-    /**
-     * @var Domain
-     */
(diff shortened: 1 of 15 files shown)


src/Model/Domain/AbstractDomainResponse.php
```diff
@@ -68,50 +45,41 @@ public static function create(array $data)
...
-     */
-    public function getDomain()
+    public function getDomain(): ?Domain
     {
         return $this->domain;
     }
 
     /**
-     * @return DnsRecord[]
+     * @return DnsRecord[] tx
```
DavidGarciaCat: Was the `tx` annotation intentional?

Nyholm: Yes, But Im not sure what "tx" is. :/

DavidGarciaCat: Maybe TXT records?

PR #542, opened 2019-01-16
Fixed PHP7 on Event model and Ip model

Nyholm:
Two more namespaces updated to PHP 7 code.


Nyholm:
We should always try to fix the issue. Phpstan helps us to detect bugs. So making the config more relaxed or optional will not help improving the code. 

The reason master is green and the PR fails due unrelated code is because phpstan released a new version. 

Ive updated the PR.

Nyholm:
Thank you for the review and for merging

PR #546, opened 2019-01-22
Updating Suppression models to php7

aradoje:
I have messed up something and created a files mess for @Nyholm to swim through :)


Nyholm:
New PR is here: #547

PR #547, opened 2019-01-22
Update Suppression models to php7

aradoje:


src/Model/Suppression/BaseResponse.php
```diff
@@ -20,49 +20,29 @@
...
     private $message;
 
-    /**
-     * @param string $address
-     * @param string $message
-     */
-    private function __construct($address, $message)
+    private function __construct(string $address, string $message)
     {
         $this->address = $address;
```
Nyholm: We can move this to ::create

src/Model/Suppression/Complaint/Complaint.php
```diff
@@ -16,57 +16,33 @@
...
-     * @return Complaint
-     */
-    public static function create(array $data)
+    public static function create(array $data): self
     {
         $complaint = new self($data['address']);
 
         if (isset($data['created_at'])) {
-            $complaint->setCreatedAt(new \DateTime($data['created_at']));
+            $complaint->createdAt = new \DateTime($data['created_at']);
```
Nyholm: Lets use `\DateTimeImmutable`

src/Model/Suppression/BaseResponse.php
aradoje: What do you mean? To have empty constructor and set the properties in create method?

Nyholm: Yeah, correct

src/Model/Suppression/BaseResponse.php
```diff
@@ -20,49 +20,28 @@
...
 
-    /**
-     * @return BaseResponse
-     */
-    public static function create(array $data)
+    public static function create(array $data): self
     {
-        $address = isset($data['address']) ? $data['address'] : '';
-        $message = isset($data['message']) ? $data['message'] : '';
+        $baseResponse = new static();
```
Nyholm: For consistency with other models, let's call this variable `$model`.

src/Model/Suppression/Bounce/Bounce.php
```diff
@@ -16,105 +16,51 @@
...
-    /**
-     * @return Bounce
-     */
-    public static function create(array $data)
+    public static function create(array $data): self
     {
-        $bounce = new self($data['address']);
+        $bounce = new self();
+        $bounce->address = $data['address'];
+        $bounce->createdAt = new \DateTimeImmutable();
```
Nyholm: We are okey with this being null if not set from APi

src/Model/Suppression/Bounce/Bounce.php
```diff
@@ -16,105 +16,51 @@
...
 
         if (isset($data['code'])) {
-            $bounce->setCode($data['code']);
+            $bounce->code = $data['code'];
```
Nyholm: Do `$model->code = $data['code'] ?? null;` and remove the if-statement

src/Model/Suppression/Complaint/Complaint.php
```diff
@@ -16,57 +16,31 @@
...
-    /**
-     * @return Complaint
-     */
-    public static function create(array $data)
+    public static function create(array $data): self
     {
-        $complaint = new self($data['address']);
+        $complaint = new self();
+        $complaint->address = $data['address'];
+        $complaint->createdAt = new \DateTimeImmutable();
```
Nyholm: This line could be removed

src/Model/Suppression/Complaint/Complaint.php
```diff
@@ -16,57 +16,31 @@
...
 
-        if (isset($data['created_at'])) {
-            $complaint->setCreatedAt(new \DateTime($data['created_at']));
-        }
+        $complaint->createdAt = isset($data['created_at']) ? new \DateTimeImmutable($data['created_at']) : new \DateTimeImmutable();
```
Nyholm: `$model->createdAt = isset($data['created_at']) ? new \DateTimeImmutable($data['created_at']) : null`

src/Model/Suppression/Bounce/Bounce.php
```diff
@@ -16,105 +16,44 @@
...
-        if (isset($data['code'])) {
-            $bounce->setCode($data['code']);
-        }
-        if (isset($data['error'])) {
-            $bounce->setError($data['error']);
-        }
-        if (isset($data['created_at'])) {
-            $bounce->setCreatedAt(new \DateTime($data['created_at']));
-        }
+        $model->address = $data['address'];
```
Nyholm: What if `address` is not set?

src/Model/Suppression/Bounce/Bounce.php
```diff
@@ -16,105 +16,44 @@
...
+    public function getAddress(): string
     {
         return $this->address;
     }
 
-    /**
-     * @return string
-     */
-    public function getCode()
+    public function getCode(): string
```
Nyholm: Must be `: ?string`

Make sure to check all of these

src/Model/Suppression/Bounce/Bounce.php
```diff
@@ -16,105 +16,44 @@
...
-    private function setError($error)
-    {
-        $this->error = $error;
-    }
-
-    /**
-     * @return \DateTime
-     */
-    public function getCreatedAt()
+    public function getCreatedAt(): \DateTimeImmutable
```
Nyholm: Same here: `: ?\DateTimeImmutable`

src/Model/Suppression/Complaint/Complaint.php
```diff
@@ -16,57 +16,30 @@
...
-    public static function create(array $data)
+    public static function create(array $data): self
     {
-        $complaint = new self($data['address']);
+        $model = new self();
 
-        if (isset($data['created_at'])) {
-            $complaint->setCreatedAt(new \DateTime($data['created_at']));
-        }
+        $model->address = $data['address'];
```
Nyholm: What if `address` is not set?

src/Model/Suppression/Complaint/Complaint.php
```diff
@@ -16,57 +16,30 @@
...
+    public function getAddress(): string
     {
         return $this->address;
     }
 
-    /**
-     * @return \DateTime
-     */
-    public function getCreatedAt()
+    public function getCreatedAt(): \DateTimeImmutable
```
Nyholm: Nullable

src/Model/Suppression/Bounce/Bounce.php
aradoje: thought it's ment to break, that address is mandatory. Will fix it

Nyholm: I dont think we should assume anything that comes from the API

Nyholm:
This feels great:
screenshot 2019-01-22 at 16 41 54

PR #548, opened 2019-01-22
Refactor Tag models to php7 code

aradoje:


src/Model/Tag/DeleteResponse.php
```diff
@@ -18,31 +18,19 @@
...
-    /**
-     * @return DeleteResponse
-     */
-    public static function create(array $data)
+    public static function create(array $data): self
     {
-        return new self($data['message']);
+        $model = new self();
+        $model->message = $data['message'] ?? '';
     }
```
Nyholm: We need to return something

PR #561, opened 2019-02-03
We dont need any arguments since we are using a config file

Nyholm:


Nyholm:
Closed in favor of #562

PR #564, opened 2019-02-03
Show that we supoprt PSR-18

Nyholm:


README.md
```diff
@@ -24,25 +24,21 @@ composer:
...
-Mailgun has a dependency on the virtual package
-[php-http/client-implementation](https://packagist.org/providers/php-http/client-implementation) which requires you to install **an** adapter, but we do not care which one. That is an implementation detail in your application. We also need **a** PSR-7 implementation and **a** message factory. 
-
-You do not have to use the `php-http/curl-client` if you do not want to. You may use the `php-http/guzzle6-adapter`. Read more about the virtual packages, why this is a good idea and about the flexibility it brings at the [HTTPlug docs](http://docs.php-http.org/en/latest/httplug/users.html).
-
 ## Usage
 
-You should always use Composer's autoloader in your application to automatically load the your dependencies. All examples below assumes you've already included this in your file:
+You should always use Composer's autoloader in your application to automatically load
+the your dependencies. All examples below assumes you've already included this in your
```
DavidGarciaCat: Maybe removing `the` or `your`, but keeping both seems wrong to me.

Also, maybe adding `the` next to `All`?

Finally, `examples` is plural so `assumes` should be `assume`

```suggestion
your dependencies. All the examples below assume you've already included this in your
```

Nyholm: Thank you. 

Also, cool. I never seen this "suggestion" feature before

Nyholm:
As you see I use buzz instead of curl-client. It is because Curl-client has not been tagged yet. 

https://packagist.org/providers/psr/http-client-implementation

DavidGarciaCat:
Merge once one CI tools return 👍 

Thanks, @Nyholm

Issue #461, opened 2018-04-12
Update Assets to validate the limits when available on endpoints


DavidGarciaCat:
As per a ticket that I opened with Mailgun team, these are the limits:
- Domains => 1000
- Events => 300
- Stats => None
- Tags => 1000
- Suppressions => 10000
- Routes => 1000
- Mailing Lists => 100

DavidGarciaCat:
@Nyholm do you think that we should add this, or allow the API itself to return the error message when needed? I guess that makes sense not to block the potential call if the API changes their limits, so this might be useless or even a blocker. Thoughts?

Nyholm:
Yeah, sure lets add these

DavidGarciaCat:
Makes sense adding these `Assert`? Should we risk ourselves to limit the user if the API changes the limits to allow more? Or less?

Nyholm:
Let's try. 
If we add them now, we can remove them without any issues. 
It is harder to add them later.

DavidGarciaCat:
👍 I can add them

PR #565, opened 2019-02-04
Mailgun-461: Validate limits

DavidGarciaCat:
This should fix https://github.com/mailgun/mailgun-php/issues/461

I've added the assets to validate the limits for pagination according to the responses that I got from Mailgun support team.

The check validates a range, to match with the existing checks that we have for suppressions, that were added during the development phase of those features.

### DIFF
--- a/src/Api/Tag.php
+++ b/src/Api/Tag.php
@@ -38,6 +38,8 @@ class Tag extends HttpApi
     public function index(string $domain, int $limit = 100)
     {
         Assert::stringNotEmpty($domain);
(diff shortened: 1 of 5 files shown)


src/Api/Tag.php
```diff
@@ -38,6 +38,8 @@ class Tag extends HttpApi
     public function index(string $domain, int $limit = 100)
     {
         Assert::stringNotEmpty($domain);
+        Assert::lessThanEq($limit, 1000);
```
Nyholm: Why no range here?

PR #567, opened 2019-02-06
Fix attachment renaming

loucou:
The key should be 'filename' instead of 'RemoteName'

### DIFF
--- a/src/Message/MessageBuilder.php
+++ b/src/Message/MessageBuilder.php
@@ -270,7 +270,7 @@ class MessageBuilder
 
         $this->message['attachment'][] = [
             'filePath' => $attachmentPath,
-            'remoteName' => $attachmentName,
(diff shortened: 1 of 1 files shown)


loucou:
Hi David,
thanks for your reply.

Here is the situation: 

In MessageBuilder.php, the new file name is assigned to the key 'remoteName':
        $this->message['attachment'][] = [
            'filePath' => $attachmentPath,
            'remoteName' => $attachmentName,
        ];

In Message.php, the new file name is taken from the 'filename' key:
$filename = isset($filePath['filename']) ? $filePath['filename'] : null;

And the travis test looks for 'remoteName'

So we could either modify the travis test and MessageBuilder to use 'filename', or we could modify Message to use 'remoteName'.

I will make a new PR with all the changes.
regards

Issue #568, opened 2019-02-09
Documentation Example Broken


jacobhaug:
So, I copied and pasted the below code exactly from the example referenced in the readme. 
<img width="953" alt="documentation" src="https://user-images.githubusercontent.com/1895442/52515947-841ab900-2bf0-11e9-99ff-826e4d8cfe5c.png">

Composer installed worked fine.
<img width="1078" alt="composer-intall" src="https://user-images.githubusercontent.com/1895442/52515950-a9a7c280-2bf0-11e9-89ac-021787e76d20.png">

**However, when loading the page I get the below fatal error.**

`Fatal error: Uncaught TypeError: Argument 1 passed to Buzz\Client\AbstractClient::__construct() must be of the type array, object given, called in /Users/jacob/Code/qc/vendor/php-http/discovery/src/Strategy/CommonClassesStrategy.php on line 119 and defined in /Users/jacob/Code/qc/vendor/kriswallsmith/buzz/lib/Client/AbstractClient.php:31 Stack trace: #0 /Users/jacob/Code/qc/vendor/php-http/discovery/src/Strategy/CommonClassesStrategy.php(119): Buzz\Client\AbstractClient->__construct(Object(Nyholm\Psr7\Factory\HttplugFactory)) #1 /Users/jacob/Code/qc/vendor/php-http/discovery/src/ClassDiscovery.php(207): Http\Discovery\Strategy\CommonClassesStrategy::buzzInstantiate() #2 /Users/jacob/Code/qc/vendor/php-http/discovery/src/HttpClientDiscovery.php(34): Http\Discovery\ClassDiscovery::instantiateClass(Array) #3 /Users/jacob/Code/qc/vendor/mailgun/mailgun-php/src/Mailgun/HttpClientConfigurator.php(160): Http\Discovery\HttpClientDiscovery::find() #4 /Users/jacob/Code/qc/vendor/mailgun/mailgun-php/src/Mailgun/HttpClientConfigurato in /Users/jacob/Code/qc/vendor/kriswallsmith/buzz/lib/Client/AbstractClient.php on line 31`

jacobhaug:
Just in case you felt this was an issue with PHP, I've tested it with PHP 7.3.1 and PHP 7.2.14.

loomsen:
Same issue here.

StoryHack:
Can verify, I'm getting the same thing.

StoryHack:
[...]

PR #575, opened 2019-02-19
Updated composer install link

samuelgfeller:
The old link gave an errror (https://github.com/mailgun/mailgun-php/issues/568) and this one works


Nyholm:
That one does not support psr18 yet. Buzz does. 

Try running the command again now when buzz 1.0 is released.

PR #618, opened 2019-08-09
MG-617: Fix URI for the Email Validation processes

### LINKED ISSUE #617: Email Validation Not Working
In 3.0.0-beta1, the email validation function fails because of an invalid response from the API. It appears that, in EmailValidation.php, the GET request on Line 54 `/address/private/validate` should be `/v4/address/validate` (and Line 93 as well).

DavidGarciaCat:
This should fix #617

- [ ] Replace URL for a single email validation
- [ ] Build new integration for bulk email validations, replacing the current code

https://documentation.mailgun.com/en/latest/api-email-validation.html#

### DIFF
--- a/src/Api/EmailValidation.php
+++ b/src/Api/EmailValidation.php
@@ -51,7 +51,7 @@ class EmailValidation extends HttpApi
             'mailbox_verification' => $mailboxVerification,
         ];
 
-        $response = $this->httpGet('/address/private/validate', $params);
+        $response = $this->httpGet('/v4/address/validate', $params);
 
         return $this->hydrateResponse($response, ValidateResponse::class);
     }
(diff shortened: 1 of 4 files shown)


DavidGarciaCat:
Due to some inconsistency between the EU API and the US API, I've raised a ticket to the Mailgun support team, to get a final confirmation about what edge cases we need to cover.

bogus-glitch:
Hello,
What is the status with this PR?  
Thanks!

DavidGarciaCat:
Hi @bogus-glitch 

After a deeper review, more or less when the SDK v3 was released, I found that this implementation is not the right one.

You still can use the old v3 validation if you use the public key and the US servers, as you can read on the issue that raised this PR: https://github.com/mailgun/mailgun-php/issues/617#issuecomment-520191517

I will work on an update of this PR as soon as I get some time. In the meantime, the solution provided by the Mailgun support team (that comment) should work.

Cheers,

PR #621, opened 2019-08-26
Add Forbidden response

Littlesqx:
I want to consider forbidden response ( etc. Account disable) instead of UnknownException, So this is a way to get it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,3 @@ composer.lock
 phpunit.phar
 phpunit.xml
 modd.conf
-.idea
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ composer.lock
 phpunit.phar
 phpunit.xml
 modd.conf
+.idea
--- a/tests/Exception/HttpClientExceptionTest.php
+++ b/tests/Exception/HttpClientExceptionTest.php
@@ -34,4 +34,30 @@ class HttpClientExceptionTest extends MailgunTestCase
         $exception = HttpClientException::badRequest($response);
         $this->assertStringEndsWith('<html><body>Server HTML</body></html>', $exception->getMessage());
     }
+
+    public function testForbiddenRequestThrowException()
+    {
+        $response = new Response(403, ['Content-Type' => 'application/json'], '{"Error":"Business Verification"}');
+        $exception = HttpClientException::forbidden($response);
+        $this->assertInstanceOf(HttpClientException::class, $exception);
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -5,3 +5,4 @@ composer.lock
 phpunit.phar
 phpunit.xml
 modd.conf
+.idea
```
DavidGarciaCat: Can you please remove this rule? Any non-related file or folder such as the ones generated by your IDE must be placed always in your global Git Ignore file instead of adding them on a project basis. If not, all projects could end adding many folders to exclude the IDE settings such as `.idea` for PHP Storm or `nbproject` for Netbeans just to add 2 examples.

Littlesqx: roger that. Thanks 😄

Littlesqx:
> Hello, @Littlesqx
> 
> Many thanks for your PR.
> 
> Although your update makes sense to me, I have two comments for this PR:
> 
> * I've added an in-line comment regarding the `.gitignore` file, please take a look there for further details
> * Would it be possible for you to add an automated test to ensure the new update works? If we don't have any other test for the rest of the status codes, don't worry about them, due to we can add them later - but all new updates should include automated tests as far as it is possible to add them.
> 
> Thanks,

I have updated the code, please help me review it. Thanks 😄

tests/Exception/HttpClientExceptionTest.php
```diff
@@ -34,4 +34,30 @@ public function testBadRequestGetMessage()
...
+
+        $response = new Response(403, ['Content-Type' => 'application/json'], '{"Message":"Business Verification"}');
+        $exception = HttpClientException::forbidden($response);
+        $this->assertStringEndsWith('{"Message":"Business Verification"}', $exception->getMessage());
+    }
+
+    public function testForbiddenRequestGetMessage()
+    {
+        $response = new Response(400, ['Content-Type' => 'text/html'], '<html><body>Forbidden</body></html>');
+        $exception = HttpClientException::badRequest($response);
```
DavidGarciaCat: Hey @Littlesqx

Are you sure that we should consider a `bad request` exception when we are testing a `forbidden` implementation?

Cheers,

Littlesqx: @DavidGarciaCat Oh, no~~~sorry for my carelessness. I didn't change the code after paste `bad request test case` 😿

DavidGarciaCat:
Thanks

PR #622, opened 2019-08-26
Revise domain creation to allow select arguments

nathanntg:
Previously, passing `$spamAction` only worked if an SMTP password (`$smtpPass`) was provided. This revision de-couples those two arguments.

In addition, it adds support for the `$wildcard` arguments (which appeared to not be used before).

Finally, this revision adds SMTP password validation (the same used in the `credentials` method).


src/Api/Domain.php
```diff
@@ -88,16 +88,26 @@ public function create(string $domain, string $smtpPass = null, string $spamActi
 
         $params['name'] = $domain;
 
-        // If at least smtpPass available, check for the fields spamAction wildcard
         if (!empty($smtpPass)) {
+            Assert::stringNotEmpty($smtpPass);
+            Assert::lengthBetween($smtpPass, 5, 32, 'SMTP password must be between 5 and 32 characters.');
```
DavidGarciaCat: Would you mind please removing this Assert? Even if at the moment the API docs are setting this requirement, we have no warranties that it will be kept in the future. It makes more sense to allow the API itself to return an error message regarding this issue so that the validation will always happen on their end.

src/Api/Domain.php
```diff
@@ -88,16 +88,26 @@ public function create(string $domain, string $smtpPass = null, string $spamActi
...
+
+            $params['smtp_password'] = $smtpPass;
+        }
+
+        if (null !== $spamAction) {
```
DavidGarciaCat: I would suggest keeping here the `!empty` approach due to:
- `''` will pass this check
- to be consistent with other checks
```suggestion
        if (!empty($spamAction)) {
```

src/Api/Domain.php
```diff
@@ -88,16 +88,26 @@ public function create(string $domain, string $smtpPass = null, string $spamActi
...
             // TODO(sean.johnson): Extended spam filter input validation.
             Assert::stringNotEmpty($spamAction);
-            Assert::boolean($wildcard);
 
-            $params['smtp_password'] = $smtpPass;
             $params['spam_action'] = $spamAction;
         }
 
+        if (null !== $wildcard) {
```
DavidGarciaCat: 👍 This one makes sense and there is no need to change it, due to the `!empty` check would not work properly for the `false` value

PR #639, opened 2019-09-18
Bug/members list

### LINKED ISSUE #637: Uncaught InvalidArgumentException: First argument to "Http\Message\MultipartStream\MultipartStreamBuilder::createStream()" must be a string, resource or StreamInterface
```
<?php

declare(strict_types=1);

require_once './vendor/autoload.php';

use Mailgun\Mailgun;

define("API_KEY", "");
define("API_DOMAIN", "");
define("SUBLIST", "");
define("EMAIL", "");

$mailgun = Mailgun::create(API_KEY);
$mailgun->mailingList()->member()->create(
    SUBLIST,
    EMAIL,
    null, // name
    [],
    false,
    true
);

$parameters = [
    'name' => null, // name
    'subscribed' => 'yes',
];

$mailgun->mailingList()->member()->update(
    SUBLIST,
    EMAIL,
    $parameters
);
```

Null name will cause error:

> PHP Fatal error:  Uncaught InvalidArgumentException: First argument to "Http\Message\MultipartStream\MultipartStreamBuilder::createStream()" must be a string, resource or StreamInterface. in /home/zolotukhin/Projects/github/test-mailgun/vendor/php-http/multipart-stream-builder/src/MultipartStreamBuilder.php:332

### LINKED ISSUE #638: Uncaught Mailgun\Exception\HttpClientException: The parameters passed to the API were invalid. Check your inputs!  'vars' parameter is not a valid JSON in vendor/mailgun/mailgun-php/src/Exception/HttpClientException.php:63
Passing `vars` parameter as array will cause error:

```
<?php

declare(strict_types=1);

require_once './vendor/autoload.php';

use Mailgun\Mailgun;

define("API_KEY", "");
define("API_DOMAIN", "");
define("SUBLIST", "");
define("EMAIL", "");
define("NAME", "");

$mailgun = Mailgun::create(API_KEY);
$mailgun->mailingList()->member()->create(
    SUBLIST,
    EMAIL,
    NAME,
    [],
    false,
    true
);

$parameters = [
    'vars' => [
        'foo' => 'bar',
    ],
    'name' => NAME,
    'subscribed' => 'yes',
];

$mailgun->mailingList()->member()->update(
    SUBLIST,
    EMAIL,
    $parameters
);
```

> PHP Fatal error:  Uncaught Mailgun\Exception\HttpClientException: The parameters passed to the API were invalid. Check your inputs!
> 
> 'vars' parameter is not a valid JSON in /home/zolotukhin/Projects/github/test-mailgun/vendor/mailgun/mailgun-php/src/Exception/HttpClientException.php:63

`vars` parameters are not json encoded ([only value](https://github.com/mailgun/mailgun-php/blob/master/src/Api/MailingList/Member.php#L202)) while request

indapublic:
Fixes #637 
Fixes #638 

Will test with my use cases today later

### DIFF
--- a/src/Api/MailingList/Member.php
+++ b/src/Api/MailingList/Member.php
@@ -156,6 +156,7 @@ class Member extends HttpApi
                     case 'vars':
                         if (is_array($value)) {
                             $value = json_encode($value);
+                            $parameters[$field] = $value;
                         }
                         // We should assert that "vars"'s $value is a string.
                         // no break
@@ -203,6 +204,7 @@ class Member extends HttpApi
                 case 'vars':
                     if (is_array($value)) {
                         $value = json_encode($value);
+                        $parameters[$field] = $value;
                     }
                     // We should assert that "vars"'s $value is a string.
                     // no break
--- a/src/Api/MailingList/Member.php
+++ b/src/Api/MailingList/Member.php
@@ -223,7 +223,7 @@ class Member extends HttpApi
             }
         }
 
-        if (null === $parameters['name']) {
+        if (array_key_exists('name', $parameters) && null === $parameters['name']) {
             unset($parameters['name']);
         }
 
--- a/src/Api/MailingList/Member.php
+++ b/src/Api/MailingList/Member.php
@@ -156,7 +156,6 @@ class Member extends HttpApi
                     case 'vars':
                         if (is_array($value)) {
                             $value = json_encode($value);
-                            $parameters[$field] = $value;
                         }
                         // We should assert that "vars"'s $value is a string.
                         // no break
@@ -199,12 +198,11 @@ class Member extends HttpApi
         Assert::stringNotEmpty($address);
         Assert::isArray($parameters);
 
-        foreach ($parameters as $field => $value) {
+        foreach ($parameters as $field => &$value) {
             switch ($field) {
                 case 'vars':
                     if (is_array($value)) {
                         $value = json_encode($value);
-                        $parameters[$field] = $value;
                     }
                     // We should assert that "vars"'s $value is a string.
                     // no break
--- a/src/Api/MailingList/Member.php
+++ b/src/Api/MailingList/Member.php
@@ -99,11 +99,14 @@ class Member extends HttpApi
 
         $params = [
             'address' => $address,
-            'name' => $name,
             'vars' => \json_encode($vars),
             'subscribed' => $subscribed ? 'yes' : 'no',
             'upsert' => $upsert ? 'yes' : 'no',
         ];
+
+        if (null !== $name) {
+            $params['name'] = $name;
+        }
 
         $response = $this->httpPost(sprintf('/v3/lists/%s/members', $list), $params);
 
@@ -204,8 +207,11 @@ class Member extends HttpApi
                     // We should assert that "vars"'s $value is a string.
                     // no break
                 case 'address':
-                case 'name':
                     Assert::stringNotEmpty($value);
+
+                    break;
+                case 'name':
+                    Assert::nullOrStringNotEmpty($value);
 
                     break;
                 case 'subscribed':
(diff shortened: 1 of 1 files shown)


src/Api/MailingList/Member.php
```diff
@@ -153,6 +156,7 @@ public function createMultiple(string $list, array $members, $upsert = false)
                     case 'vars':
                         if (is_array($value)) {
                             $value = json_encode($value);
+                            $parameters[$field] = $value;
```
palpalani: i hope, `$parameters[$field] = $value;` should be `$members[$field] = $value`;

indapublic: Line removed as not wrong and not needed anymore. Because passing by reference is using here. Passing by value in `update` function was replaced

src/Api/MailingList/Member.php
```diff
@@ -195,7 +198,7 @@ public function update(string $list, string $address, array $parameters = [])
         Assert::stringNotEmpty($address);
         Assert::isArray($parameters);
 
-        foreach ($parameters as $field => $value) {
+        foreach ($parameters as $field => &$value) {
```
DavidGarciaCat: Why adding the `&` here? I can't see any need for this

src/Api/MailingList/Member.php
```diff
@@ -204,9 +207,12 @@ public function update(string $list, string $address, array $parameters = [])
                     // We should assert that "vars"'s $value is a string.
                     // no break
                 case 'address':
-                case 'name':
                     Assert::stringNotEmpty($value);
 
+                    break;
+                case 'name':
+                    Assert::nullOrStringNotEmpty($value);
```
DavidGarciaCat: Based on the condition below...
```php
        if (array_key_exists('name', $parameters) && null === $parameters['name']) {
            unset($parameters['name']);
        }
```
don't you think that's better to check if it's a non-empty string instead of allow `null` and then unset the key?

indapublic: Null is allowed value for member name in Mailgun

src/Api/MailingList/Member.php
indapublic: We are [encoding](https://github.com/mailgun/mailgun-php/blob/master/src/Api/MailingList/Member.php#L202) vars parameters but original value remains the same. It will cause exception #638 if `vars` will be passed like array

indapublic: We are using `&` [here](https://github.com/mailgun/mailgun-php/blob/master/src/Api/MailingList/Member.php#L147)

indapublic:
Hi, @DavidGarciaCat . Do this PR have any chance to be approved?

src/Api/MailingList/Member.php
DavidGarciaCat: I don't like to pass variables by reference (adding the `&`) unless it is strictly necessary, and I am not sure that this is the case here.

However, I am happy to keep it as far as it doesn't have any direct impact when managing the data in `$value`.

indapublic:
Hi @DavidGarciaCat Thanks for response
But you using pass by reference [already](https://github.com/mailgun/mailgun-php/blob/master/src/Api/MailingList/Member.php#L147), I just repeat this behaviour

DavidGarciaCat:
There're 3 errors when running PHPStan but they are not related to this update. We might need to fix these ones first and then re-run the tests again. Something that shouldn't take too much time.

If you want to expedite this update, feel free to add one more commit with the patch and I can take care of it. If not, I will need to work on the patch first and then rerun the CI for this PR.

Thanks for your contribution!

src/Api/MailingList/Member.php
```diff
@@ -99,12 +99,15 @@ public function create(string $list, string $address, string $name = null, array
...
             'address' => $address,
-            'name' => $name,
             'vars' => \json_encode($vars),
             'subscribed' => $subscribed ? 'yes' : 'no',
             'upsert' => $upsert ? 'yes' : 'no',
         ];
 
+        if (null !== $name) {
+            $params['name'] = $name;
+        }
```
setkyar, line 109:
```
!is_null($name)
```

I thinks `is_null` function is more readable.

DavidGarciaCat (approved):

PR #643, opened 2019-10-03
Add unsubscribe delete API with parameter

iwahara:
ref https://documentation.mailgun.com/en/latest/api-suppressions.html#delete-a-single-unsubscribe


iwahara:
Hi @DavidGarciaCat 

thanks to merge!  because I use this lib too and I was very helpful.

PR #653, opened 2019-11-22
Cast route `priority` param to string (fix #652)

### LINKED ISSUE #652: Broken $client->routes()->create() on v2.8.1
Hi,

I've found that on V2.8.1 it's impossible to create a route due to an internal SDK error.

```
    $client->routes()->create([
        "expression" => "match_recipient('abc@xyz.com')",
        "actions" => "store(notify='https://mycallbackurl.com')",
        "description" => "foo bar biz"
    ]);
```

This will always throw an exception:

```
In MultipartStreamBuilder.php line 332:
                                                                                                                                          
  [InvalidArgumentException]                                                                                                              
  First argument to "Http\Message\MultipartStream\MultipartStreamBuilder::createStream()" must be a string, resource or StreamInterface.  
                                                                                                                                          

Exception trace:
  at /var/www/vendor/php-http/multipart-stream-builder/src/MultipartStreamBuilder.php:332
 Http\Message\MultipartStream\MultipartStreamBuilder->createStream() at /var/www/vendor/php-http/multipart-stream-builder/src/MultipartStreamBuilder.php:90
 Http\Message\MultipartStream\MultipartStreamBuilder->addResource() at /var/www/vendor/mailgun/mailgun-php/src/Mailgun/RequestBuilder.php:62
 Mailgun\RequestBuilder->create() at /var/www/vendor/mailgun/mailgun-php/src/Mailgun/Api/HttpApi.php:160
 Mailgun\Api\HttpApi->httpPostRaw() at /var/www/vendor/mailgun/mailgun-php/src/Mailgun/Api/HttpApi.php:144
 Mailgun\Api\HttpApi->httpPost() at /var/www/vendor/mailgun/mailgun-php/src/Mailgun/Api/Route.php:91
 Mailgun\Api\Route->create() at /var/www/my-script.php:60
```

iget-master:
This PR fixes the issue #652 by casting the `priority` parameter to string before calling `httpPost` method.

This fix is for `v2.x`, I'm not sure if v3.0.0 have same issue.

I really suggest this fix to be merged immediately and released a patch version containing it, since the last v2.8.1 is broken.

### DIFF
--- a/src/Mailgun/Api/Route.php
+++ b/src/Mailgun/Api/Route.php
@@ -82,7 +82,7 @@ class Route extends HttpApi
         Assert::integer($priority);
 
         $params = [
-            'priority' => $priority,
+            'priority' => (string) $priority,
             'expression' => $expression,
             'action' => $actions,
             'description' => $description,
--- a/tests/Api/RouteTest.php
+++ b/tests/Api/RouteTest.php
@@ -47,7 +47,7 @@ class RouteTest extends TestCase
         $api = $this->getApiMock();
         $api->expects($this->once())
             ->method('httpPost')
-            ->with([
+            ->with('/v3/routes', [
                 'priority' => '100',
                 'expression' => 'catch_all()',
                 'action' => ['forward("mailbox@myapp.com")'],
--- a/tests/Api/RouteTest.php
+++ b/tests/Api/RouteTest.php
@@ -47,6 +47,12 @@ class RouteTest extends TestCase
         $api = $this->getApiMock();
         $api->expects($this->once())
             ->method('httpPost')
+            ->with([
+                'priority' => '100',
+                'expression' => 'catch_all()',
+                'action' => ['forward("mailbox@myapp.com")'],
+                'description' => 'example'
+            ])
             ->willReturn(new Response());
 
         $api->create('catch_all()', ['forward("mailbox@myapp.com")'], 'example', 100);


iget-master:
I've also updated the RouteTest.php to assert that the priority parameter is casted corretly.

kslr:
No one is responsible for the project?

iget-master:
@DavidGarciaCat can you please give some attention to this? The PR is ready for merge (not passing in your CI because your CI is broken)

DavidGarciaCat:
Hi @iget-master 

Personally, knowing that the SDK v2 is deprecated and replaced by v3 due to:
- The support for PHP 5.5, that is too old
- The obsolete packages, that they have been replaced
- The newer releases

I don't think that this PR makes sense anymore.

Keep adding more and more updates here will lead to the need to maintain two versions, when v3 was released aiming to replace the old v2. I really hope that it makes sense.

If the problem persists on v3 then I am more than happy to consider the update targeting the master branch, as far as all CI tools are passing.

Cheers,

iget-esoares:
But v3 migrating is impossible here due to package incompatibility. Also, v2 used to work in previous versions, and was broken. Why don't release a patch to v2? Really see no reason to don't do it. It's not a feature or a breaking change, it's a bug fix!

DavidGarciaCat:
According to the Mailgun documentation (not attached to any programming language or a specific client), the `priority` is an integer - not a string - with this description:

> Integer: smaller number indicates higher priority.
> Higher priority routes are handled first. Defaults to 0.

This is why I am not sure about this particular PR.

Can you please paste here somewhere the list of the dependencies that you are using so that I can try to replicate the error? If the only way to fix it is by merging this PR, I'm happy to do it as far as the CI tools report thumbs up, we shouldn't break the analysis of the code by adding a patch.

Hope it makes sense to you.

iget-esoares:
Yes, the API expects an integer, but you agree that HTTP protocol only receives strings? (Even JSON ate strings) The HTTP client expects a string, but receives an integer, and fails. In the end, the API will still receive a integer (encoded in a string).

Sorry for the explanation, but I'm on the mobile right now.

If you just try to use the route()->create() method you will notice that it fails if you pass a integer, and fail if you pass a string (it validates if the priority field is integer)

iget-master:
@DavidGarciaCat if you check the Issue #652 (precisely [this comment](https://github.com/mailgun/mailgun-php/issues/652#issuecomment-557534697)) you will understand why it must be a string.

tomschlick:
Hoping to see this get merged soon for the `2.x` branch. For my own projects I can't upgrade to `3.x` yet because of other dependencies that haven't upgraded to `php-http/httplug@2.0` yet. 

This would provide relief from the workaround @iget-master had to put together [here](https://github.com/mailgun/mailgun-php/issues/652#issuecomment-567190676), especially since its a pretty critical bug that wont let you create anything that contains a priority.

DavidGarciaCat:
Time has passed, and I can see the Mailgun Documentation was not yet updated. I have sent a request to their support team. As soon as I get a reply from them, I'll update this PR accordingly.

DavidGarciaCat:
By the way: I won't be able to merge this update unless Travis CI returns 👍 so maybe we should consider an update if this is still relevant

kslr:
Good news for today

DavidGarciaCat:
According to Mailgun Support, `integer` should be valid and working. I am sharing with you (below) a screenshot with the reply from the support team.

If this still throws an error, can someone please provide me with an example with a failure, so that I can send it back?

image

DavidGarciaCat:
Mailgun Support team has reached me again.

Given I didn't get a failing example to prove the error and given the functional API calls that they provided (that I attached five days ago), the support case is now resolved.

Thus, there's no need to keep this Pull Request open, as it was proved not to be a requirement, and (specially) because none of the stakeholders of this update posted a failing scenario to follow up this case with their support team.

Integers seem to be a functional case, and there's no need to force them as a string.

I am closing both the Issue and the Pull Request.

PR #660, opened 2020-01-01
fix(#642) HttpApi: Expect client to be of type ClientInterface|PluginClient

tonythomas01:
Rest of the software assumes that $httpClient could be an instance of both `ClientInterface|PluginClient`. The issue was raising exceptions on the latest versions as per:
```
composer require mailgun/mailgun-php kriswallsmith/buzz nyholm/psr7
```

Thrown exceptions were similar to the one you see in the bug report. 

Happens with buzz adapter, btw.

### DIFF
--- a/src/Api/HttpApi.php
+++ b/src/Api/HttpApi.php
@@ -12,16 +12,15 @@ declare(strict_types=1);
 namespace Mailgun\Api;
 
 use Http\Client\Common\PluginClient;
-use Mailgun\Exception\UnknownErrorException;
-use Mailgun\Hydrator\Hydrator;
-use Mailgun\Hydrator\NoopHydrator;
 use Mailgun\Exception\HttpClientException;
 use Mailgun\Exception\HttpServerException;
+use Mailgun\Exception\UnknownErrorException;
 use Mailgun\HttpClient\RequestBuilder;
+use Mailgun\Hydrator\Hydrator;
+use Mailgun\Hydrator\NoopHydrator;
 use Psr\Http\Client as Psr18;
 use Psr\Http\Client\ClientInterface;
 use Psr\Http\Message\ResponseInterface;
-
 
 /**
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
@@ -47,9 +46,9 @@ abstract class HttpApi
 
     public function __construct($httpClient, RequestBuilder $requestBuilder, Hydrator $hydrator)
     {
-        if ( !$httpClient instanceof ClientInterface && !$httpClient instanceof PluginClient ) {
+        if (!$httpClient instanceof ClientInterface && !$httpClient instanceof PluginClient) {
             throw new \RuntimeException('httpClient must be an instance of
-			Psr\Http\Client\ClientInterface or Http\Client\Common\PluginClient' );
+			Psr\Http\Client\ClientInterface or Http\Client\Common\PluginClient');
         }
         $this->httpClient = $httpClient;
         $this->requestBuilder = $requestBuilder;
@@ -85,22 +84,22 @@ abstract class HttpApi
     {
         $statusCode = $response->getStatusCode();
         switch ($statusCode) {
-            case 400:
-                throw HttpClientException::badRequest($response);
-            case 401:
-                throw HttpClientException::unauthorized($response);
-            case 402:
-                throw HttpClientException::requestFailed($response);
-            case 403:
-                throw  HttpClientException::forbidden($response);
-            case 404:
-                throw HttpClientException::notFound($response);
-            case 413:
-                throw HttpClientException::payloadTooLarge($response);
-            case 500 <= $statusCode:
-                throw HttpServerException::serverError($statusCode);
-            default:
-                throw new UnknownErrorException();
+        case 400:
+            throw HttpClientException::badRequest($response);
+        case 401:
+            throw HttpClientException::unauthorized($response);
+        case 402:
+            throw HttpClientException::requestFailed($response);
+        case 403:
+            throw  HttpClientException::forbidden($response);
+        case 404:
+            throw HttpClientException::notFound($response);
+        case 413:
+            throw HttpClientException::payloadTooLarge($response);
+        case 500 <= $statusCode:
+            throw HttpServerException::serverError($statusCode);
+        default:
+            throw new UnknownErrorException();
         }
     }
 
--- a/src/Api/HttpApi.php
+++ b/src/Api/HttpApi.php
@@ -11,6 +11,7 @@ declare(strict_types=1);
 
 namespace Mailgun\Api;
 
+use Http\Client\Common\PluginClient;
 use Mailgun\Exception\UnknownErrorException;
 use Mailgun\Hydrator\Hydrator;
 use Mailgun\Hydrator\NoopHydrator;
@@ -21,6 +22,7 @@ use Psr\Http\Client as Psr18;
 use Psr\Http\Client\ClientInterface;
 use Psr\Http\Message\ResponseInterface;
 
+
 /**
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
  */
@@ -29,7 +31,7 @@ abstract class HttpApi
     /**
      * The HTTP client.
      *
-     * @var ClientInterface
+     * @var ClientInterface|PluginClient
      */
     protected $httpClient;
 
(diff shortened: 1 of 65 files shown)


tonythomas01:
I wonder if those php stan complaints are related to the change though ...

src/Api/HttpApi.php
```diff
@@ -29,7 +30,7 @@ abstract class HttpApi
     /**
      * The HTTP client.
      *
-     * @var ClientInterface
+     * @var ClientInterface|PluginClient
```
DavidGarciaCat: What is the vendor of this Client?

```
use Http\Client\Common\PluginClient;
```

The HTTP-Plug provides the `Psr\Http\...` namespace, and I don't think that is a good idea to unbind the HttPlug package or keep the constructor open to other clients that don't follow the PSR implementation.

Remove the type hint as part of the constructor looks wrong to me, especially because PHP 7 has been introducing more and more type hints everywhere.

DavidGarciaCat:
Closing due to inactivity

llbbl:
Unrelated to what Tony is working on, we are building a Laravel Nova Mailgun package and this PR solves the following error (#642):

```
Uncaught TypeError: Argument 1 passed to Mailgun\Api\HttpApi::__construct() must implement interface Psr\Http\Client\ClientInterface
```

After reviewing the changes it seems like a good PR @DavidGarciaCat . I would ask that you please consider reopening or allowing some version of these changes into the library. 

Specifically removing the ClientInterface from the constructor is what solves it. We believe the changes @tonythomas01 came up with to be a good approach to optionally passing in the ClientInterface while still allowing for the Exception to be throwing if the first parameter doesn't match what is expected. 

As it stands now the mailgun-php library is not usable at the 3.0.0 tag. 

Our options for solving this are: 

1. Keep trying older tags until we get a stable version of the library
2. Modify files in vendor 
3. Wait until this PR is resolved
4. Fork

Using Guzzle over Buzz DOES NOT solve this exception for us, but this PR changes DOES work for either Guzzle or Buzz. 

```
kriswallsmith/buzz nyholm/psr7
```
or

```
       "mailgun/mailgun-php": "^3.0",
        "guzzlehttp/guzzle": "^6.3",
        "php-http/guzzle6-adapter": "^2.0"
```

DavidGarciaCat:
Hi @llbbl 

I'm going to ask the same question that I didn't get an answer: https://github.com/mailgun/mailgun-php/pull/660#discussion_r367157248

> What is the vendor of this Client?
>
>```
>use Http\Client\Common\PluginClient;
>```
>
>The HTTP-Plug provides the `Psr\Http\...` namespace, and I don't think that is a good idea to unbind the HttPlug package or keep the constructor open to other clients that don't follow the PSR implementation.
>
>Remove the type hint as part of the constructor looks wrong to me, especially because PHP 7 has been introducing more and more type hints everywhere.

If we can clarify this, I'm more than happy to re-open and review it again.

tonythomas01:
Sorry for the delay people. I will look into it today and try to reproduce what I was dealing with.

llbbl:
Thanks for reconsidering this PR @DavidGarciaCat ! 

It's not so much as Removing the type hint, as MOVING it from constructor into the phpdoc block. Moving the type hinting into @param does not break PhpStorm and is fully supported by the Phpdoc standard. 

Furthermore, there are some other classes that use it. 
https://github.com/php-http/client-common/blob/2.3.0/src/PluginClient.php

or here is good one from symfoy-process
https://github.com/symfony/process/blob/v5.1.5/Pipes/AbstractPipes.php#L33

Next item you mention; The vendor of

> Http\Client\Common\PluginClient

is included from:

vendor/php-http/client-common/src/PluginClient.php

Which is pulled from this library itself here (from our composer.lock file):

```
"name": "php-http/client-common",
            "version": "1.10.0",
```
via

```
mailgun-php composer.lock:
                "php-http/client-common": "^1.9 || ^2.0",
```

After looking at the Packagist for php-http/client-common this morning, it seems probably a good idea to update the mailgun-php library composer file to a more current version. In version 1.10 they still support php 5.4 but dropped it in version 2. 

https://packagist.org/packages/php-http/client-common#2.3.0

The current version of 2.3 seems would be good, but I'll leave that to you how to handle.

llbbl:
Don't worry @tonythomas01 I got this :P

DavidGarciaCat:
Hi, @llbbl and @tonythomas01 

Thanks for the additional information. I guess that makes sense apply a few additional changes to make the system more flexible and with updated dependencies (replacing an old dependency by a new one that no longer supports PHP 5, as the SDK already does).

If Tony can send an update with the new points, we can review the changes and proceed.

I'm reopening this PR.

Cheers,

src/Api/HttpApi.php
tonythomas01: It is sad that the current verison of PHP still do not support union of types yet. However, I see the next release on Nov 2020 will have it! Yay. https://stackoverflow.com/a/58826132 

Until then, I think we have to do the manual workaround I have down here in the constructor ? I can add a @FIXME or something to remind later.

DavidGarciaCat: I guess that we can keep the current conditional to ensure we don't inject anything unexpected. Maybe the only relevant comment is to write the Exception message in a single line rather than multiline, but that's a minor point.

src/Message/MessageBuilder.php
```diff
@@ -85,8 +85,8 @@ private function getFullName(array $params): string
 
     /**
      *     @var string
-     *     @var string $first
-     *     @var string $last
+     *     @var string
+     *     @var string
```
DavidGarciaCat, line 89: These annotations are for

```php
parseAddress(string $address, array $variables): string
```

So they don't seem to be consistent. Can you please address this?

DavidGarciaCat (commented):
I can see that we're moving forward. Good!

There's still a failure on Travis CI that won't allow me to merge the new code. It should be addressed too.

src/Message/MessageBuilder.php
tonythomas01: oh yes. This should be: 
```
    /**
     * @param array $variables {
     *
     *     @var string $full_name
     *     @var string $first
     *     @var string $last
     * }
     */
```

llbbl:
Good progress yesterday!

src/Api/HttpApi.php
tonythomas01: Yeah. I think it was `pretty fix` doing that magic there!

llbbl:
Well we got the Laravel Nova Package working well! Just waiting on this. :smile:

llbbl:
Any updates on this ? :smile: @DavidGarciaCat @tonythomas01

tonythomas01:
From what I see, we dont have any failing tests at this point. So if the code looks good, it should be ready to ship.

llbbl:
@DavidGarciaCat Can we please get this merged and new version tagged?

DavidGarciaCat:
Hello there,

@llbbl apparently, this branch has one or more conflicts, and I am not allowed to merge it until these are fixed. I believe Tony should pull the latest changes on master and either rebase or merge to solve the conflicts before moving forward.

@tonybolzan as a minor comment, I would prefer (if you don't mind) use the `ClassName::class` approach instead of hardcoding the string when running the conditional for the `is_a` and throw an exception when the provided class doesn't match with either `ClientInterface` or `PluginClient` as it will keep consistency with other pieces of code under the `src` folder.

Cheers,

tonybolzan:
I believe you wanted to mention @tonythomas01

DavidGarciaCat:
My bad - sorry for the confusion

llbbl:
@tonythomas01 Have some time to address the merge conflicts or want to add me to your fork? I can help in the next week get this wrapped up.

DavidGarciaCat:
I would like to see this PR working, so it can be included as part of the next release (maybe later this week?) depending on the other two PRs that were sent and reviewed recently.

Also, it's worth mentioning that the PR https://github.com/mailgun/mailgun-php/pull/711 updates the `Domain` class again, so maybe we should coordinate the updates to avoid unnecessary commits to fix this PR if it implies a new set of errors later.

Let's check if they can cause another conflict, and let's get it ready!

llbbl:
I guess I can submit a new PR from a new fork if Tony doesn't get back to us quick enough. Glad to hear you are on board with getting these PRs wrapped up in next week David.

tonythomas01:
Sorry guys. Got caught up. Adding it to my TODO for the day.

DavidGarciaCat:
Hello @tonythomas01 

I'm really sorry to bother you, but although the TravisCI integration confirms all tests are passing, I still can see the following message even with the last merge (150ae8c) from `master` to your branch.

```
This branch cannot be rebased due to conflicts

Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch.
```

If you have a chance, can you please retake a look, so we can merge it and tag a new version?

Thank you very much for your time and patience

tonythomas01:
Hmm. Interesting. I merged upstream:master on this branch again. Lets see if this one works.

DavidGarciaCat:
:thinking: now I find it interesting...

By default, Github sets `Rebase and merge` for me, and it means I can't do it because I still get the same error:

image

However, it looks like I can change the merge process to `Squash and merge`, and then yes, it seems to be working:

image

OK, I will try to take a look later again, just to confirm all the code makes sense (as it should, as per previous reviews) and merge.

Thanks for your time, and sorry for keep bothering you.

PR #664, opened 2020-02-09
Bug fix with getContentIdMap return type

lvdhoorn:
The response from the api is an array and not a string.


DavidGarciaCat (approved):
Missing test, but we don't have a Test class and this is definitively fixing a bug, as per the API docs:

https://documentation.mailgun.com/en/latest/api-sending.html#retrieving-stored-messages

I have relaunched a rebuild of the PR because Travis CI was flagging it as a failure. If it works, I am happy to merge it.

PR #668, opened 2020-02-26
fix /v3/domains/{domain}/ips endpoint

tomschlick:
`/v3/domains/{domain}/ips` is the correct endpoint per https://documentation.mailgun.com/en/latest/api-ips.html#ips


DavidGarciaCat (approved):
Missing tests, but we don't have an `IpTest` class. Something that we should add...

PR #675, opened 2020-03-30
feat: add array support to webhook handling

martin-sanjuan:
Change webhook create() $url parameter from string to array to support
multiple URLs
Change webhook update() $url parameter from string to array yo support
multiple URLs


DavidGarciaCat:
I have re-queued the build due to the failing Travis CI report

Also, this update seems to be a duplicated case of https://github.com/mailgun/mailgun-php/pull/609 - Further investigation before merging is required

PR #680, opened 2020-05-06
Allow php-http/guzzle6-adapter ^2.0

boboldehampsink:
It still uses the same API, just updated for PHP 7


DavidGarciaCat (approved):
Thanks :-)

PR #684, opened 2020-06-08
Added Support for Mailing List "reply_preference" parameter

twoonesixdigital:
## Problem
When creating new Mailing Lists via the SDK, it's not possible to explicitly set the 'reply_preference' parameter. As a result, lists created via the SDK will always default to reply_preference = list.

## Changes
I've updated the following classes to add support for this parameter when creating new mailing lists:
- Mailgun\Model\MailingList\MailingList
- Mailgun\Api\MailingList
- Mailgun\Tests\Api\MailingListTest

## Remaining Issues
This PR doesn't extend to the update method, although I suppose it would be near trivial to make the necessary changes to the API class and its test class.

### DIFF
--- a/src/Api/MailingList.php
+++ b/src/Api/MailingList.php
@@ -63,18 +63,20 @@ class MailingList extends HttpApi
      *
      * @throws \Exception
      */
-    public function create(string $address, string $name = null, string $description = null, string $accessLevel = 'readonly')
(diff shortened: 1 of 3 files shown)


DavidGarciaCat:
Hi @twoonesixdigital 

We need these CI tools passing before merging, can you please fix the PR to solve this?

Thanks,

DavidGarciaCat:
Thanks - I'll send another PR soon with the CI fixes

PR #687, opened 2020-06-25
Fix misspelled parameter

shaggy8871:
The ips()->assign() method fails because the required parameter is misspelled. This appears to be in all releases, not just the 2.x branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ cache:
         - $HOME/.composer/cache/files
 
 php:
-  - 5.5
   - 5.6
   - 7.0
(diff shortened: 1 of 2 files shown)


shaggy8871:
@DavidGarciaCat Done. I had to remove hhvm because it's no longer supported (see https://github.com/composer/composer/issues/8127).

DavidGarciaCat (approved):
Thanks

DavidGarciaCat:
Please use the commit ID while a new version is not tagged. I will need to add some release notes and fix the CI too before that.

PR #688, opened 2020-07-07
Fix link to documentation

Casmo:
Fix the link to the api reference.


DavidGarciaCat:
I won't be able to merge unless the CI tools return thumbs up. Id you don't mind, could you fix your PR so it can be done? Cheers,

Casmo:
I've no idea how to fix the "PHPStan" error(s). It has nothing to do with the change in the documentation. All other runs are green.

PR #695, opened 2020-08-20
remove deprecated pacakge from suggestions

tomschlick:
https://github.com/Bogardo/Mailgun is archived


DavidGarciaCat:
Hi @tomschlick 

Is this Laravel integration still working, although it's archived?

Cheers,

tomschlick:
@DavidGarciaCat it hasn't been updated in 11 months and doesn't work with the most recent Laravel version (7.0) and the next one that drops this month (8.0). It should be assumed broken and not suggested as a library for new users to utilize.

DavidGarciaCat (approved):
> doesn't work with the most recent Laravel version (7.0) and the next one that drops this month (8.0)

I believe these two are good points.

Issue #697, opened 2020-09-09
Not receiving Events Info


karquieta:
Hello, i'm traying to get info about the sending logs from my mailgun domain but when I checked the $result content, it is empty. The API KEY and the domain are correct, the email I'm using is also listing logs in the mailgun console.

Here my code.

            // Instantiate the client.
            $mgClient    = Mailgun::create("API KEY");
            $domain      = "domain.com";
            $queryString = array(
                "begin"        => "Wed, 1 Jan 2020 09:00:00 -0000",
                "ascending"    => "yes",
                "limit"        =>  25,
                "pretty"       => "yes",
                "recipient"    => "email@example.com"
            );

            // Issue the call to the client.
            $result = $mgClient->events()->get($domain, $queryString);

Edit: I added to $result the method getItems(); So it can be seen like this

            $result = $mgClient->events()->get($domain, $queryString)->getItems();

But, same thing, when I see the $result content I can see the quantity of events that match that email, but are empy objects.

           data: [
                 0: {}
                 1: {}
                 ....
           ]

DavidGarciaCat:
Hi there,

Can you please check the returned response using the `ArrayHydrator` as per the Readme example?

https://github.com/mailgun/mailgun-php#response

Cheers,

DavidGarciaCat:
Hi @karquieta 

Did you try setting the `ArrayHydrator`? Did it solve your problem?

Cheers,

Nyholm:
Closing due inactivity.

PR #701, opened 2020-09-25
json callback of a model for quick serialization

uavn:
Issue #697  https://github.com/mailgun/mailgun-php/issues/697

```
$mg = Mailgun::create($key);

$queryString = array(
    "begin" => "Wed, 1 Jan 2020 09:00:00 -0000",
    "ascending" => "yes",
    "limit" =>  25,
    "pretty" => "yes",
    "recipient" => "artbonvic@gmail.com"
);

$result = $mg->events()->get($domain, $queryString)->getItems();
print_r(json_encode($result, JSON_PRETTY_PRINT));
```

```
[
    {
        "event": "accepted",
        "id": "q3h-tojUSHOgasy70zJ3wA",
        "timestamp": 1601050399,
        "eventDate": {
            "date": "2020-09-25 16:13:19.000000",
            "timezone_type": 3,
            "timezone": "UTC"
        },
        "tags": [],
        "url": "",
        "severity": "",
        "envelope": {
            "sender": "postmaster@sandbox152fb160643f41f9b09b52f7b5e370ec.mailgun.org",
            "transport": "smtp",
            "targets": "artbonvic@gmail.com"
        },
        "deliveryStatus": [],
        "campaigns": [],
        "ip": "",
        "clientInfo": [],
        "reason": "",
        "userVariables": [],
        "flags": {
            "is-authenticated": true,
            "is-test-mode": false
        },
        "routes": [],
        "message": {
            "headers": {
                "to": "artbonvic@gmail.com",
                "message-id": "20200925161319.1.812E494DE2D30C2C@sandbox152fb160643f41f9b09b52f7b5e370ec.mailgun.org",
                "from": "artbonvic@gmail.com",
                "subject": "Hello"
            },
            "size": 459
        },
        "recipient": "artbonvic@gmail.com",
        "geolocation": [],
        "storage": {
            "url": "https:\/\/sw.api.mailgun.net\/v3\/domains\/sandbox152fb160643f41f9b09b52f7b5e370ec.mailgun.org\/messages\/AgEFG3zjAp2r5Ly5SHFCKoYdZyeO3zFmZA==",
            "key": "AgEFG3zjAp2r5Ly5SHFCKoYdZyeO3zFmZA=="
        },
        "method": "http"
    },
    ...
]

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,8 @@
         "php-http/multipart-stream-builder": "^1.0",
         "php-http/client-common": "^1.9 || ^2.0",
         "php-http/discovery": "^1.6",
-        "webmozart/assert": "^1.2"
+        "webmozart/assert": "^1.2",
+        "ext-json": "^1.6"
     },
     "require-dev": {
         "phpunit/phpunit": "^7.5",
--- a/src/Model/Event/Event.php
+++ b/src/Model/Event/Event.php
@@ -11,10 +11,12 @@ declare(strict_types=1);
 
 namespace Mailgun\Model\Event;
 
+use JsonSerializable;
+
 /**
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
  */
-final class Event
+final class Event implements JsonSerializable
 {
     private $event;
     private $id;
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by DavidGarciaCat** [comment]
I would definitively add a `JsonHydrator` here (https://github.com/mailgun/mailgun-php/tree/master/src/Hydrator) so that we can pass it when instantiating the client and it can return the same content that the API itself returns. In this case, there's no need for us to convert the data to another format to then serialise it again to JSON

**Comment by uavn** [comment]
> it when instantiating the client and it can return the same content that the API itself returns. In th

But what if we want use models and then, let's say, we need to return result to the browser, in this case we can just json encode model, because if we use JsonHydrator we loose advantages from using models.

The same can be done in other models as well

**Comment by DavidGarciaCat** [comment]
If your concern is keeping the object but also provide the JSON string, then IMHO we should not move forward implementing the `JsonSerializable` interface, that means additional maintenance work (and we are already a bit behind).

My suggestion is to move forward with the `JsonHydrator` as it will keep a consistent way to retrieve the responses, no matter what and without continuous maintenance on every single model of the SDK:

```php
// Using the new JsonHydrator (that we still need to add)

require 'vendor/autoload.php';

use Mailgun\HttpClient\HttpClientConfigurator;
use Mailgun\Hydrator\JsonHydrator;
use Mailgun\Mailgun;

$configurator = new HttpClientConfigurator();
$configurator->setApiKey('key-example');

$mailgun= new Mailgun($configurator, new JsonHydrator());

$jsonString = $mailgun->messages()->send('domain', [...]);
```

Alternatively, if you want to keep the Models at all cost, you can consider one of the following scenarios:

```php
// Keeping the Model

require 'vendor/autoload.php';

use Mailgun\Mailgun;

$mailgun= Mailgun::create('key-example');

$objectResponse= $mailgun->messages()->send('domain', [...]);

$jsonString = json_encode((array) $objectResponse);
```

```php
// Getting an Array

require 'vendor/autoload.php';

use Mailgun\HttpClient\HttpClientConfigurator;
use Mailgun\Hydrator\ArrayHydrator;
use Mailgun\Mailgun;

$configurator = new HttpClientConfigurator();
$configurator->setApiKey('key-example');

$mailgun= new Mailgun($configurator, new ArrayHydrator());

$arrayResponse = $mailgun->messages()->send('domain', [...]);

$jsonString = json_encode($arrayResponse);
```

But, as I said, implementing the `JsonSerializable` interface for just one case (or all the cases) also implies constant maintenance of every single Model.

Issue #677, opened 2020-04-01
Mailgun\Model\Event omits data


ser-gi-o:
Data like "log-level" is omitted when event is an object but returned correctly when requested with ArrayHydrator and NoopHydrator.  This happens because when instantiated, data is individually assigned to each class property in create method.  Property logLevel does not exist in Mailgun\Model\Event class.

```
//omits log-level, recipient-domain
$mg = Mailgun::create($apiKey); 
$response = $mg->events()->get($domain, $params);
dump($response);

//includes log-level, recipient-domain
$configurator = new HttpClientConfigurator();
$configurator->setApiKey($apiKey);
$mg = new Mailgun($configurator, new ArrayHydrator());
$response = $mg->events()->get($domain, $params);
dump($response);
```

My work around to get all the data and multiple pages:
Instantiate Mailgun with ArrayHydrator get the first set of data and then use guzzle to get the next set of paging data using $response['paging']['next'] .  

I did it this way because if you return the response as an array you cannot use 
$mg->events()->nextPage($response) function. 

```
do {
    //next page
    $response = $guzzleClient->request(
        'GET',
        str_replace('https://', 'https://api:' . $apiKey . '@', $content['paging']['next'])
     );

    $content = json_decode($response->getBody()->getContents(), true);
    $hasItems = isset($content['items']) && sizeof($content['items']);
} while ($hasItems);
```

sykezz:
I had the same paging issue with ArrayHydrator. Since I wanted the pagination, I didn't use the ArrayHydrator, but I can't access the keys in the content since the Class properties are private, so I ended up casting the array and parsing it.

DavidGarciaCat:
The `ArrayHydrator` should solve any problem with the unmapped properties (meaning that Mailgun API now returns data that didn't exist when the SDK was updated).

If anyone can send a PR with the required fix, I'm more than happy to review and merge the update.

Cheers,

DavidGarciaCat:
This issue should now be fixed as part of PR https://github.com/mailgun/mailgun-php/pull/704

There was also [this proposal](https://github.com/mailgun/mailgun-php/issues/677#issuecomment-699070759) as an alternative workaround while the issue was being fixed.

If anyone experiences further issues, please open a new Issue.

Closing this one now.

PR #704, opened 2020-10-05
added unmapped proeprty logLevel to Event model

uavn:
https://github.com/mailgun/mailgun-php/issues/677

Support **log-level** property for the Event model


DavidGarciaCat (approved):
Thanks

PR #707, opened 2020-10-05
Change Ip::index() default to return all ips instead of shared ips

josephshanak:
The GET /ips endpoint returns all IPs by default, shared ips if the dedicated parameter is false and dedicated ips only if the dedicated parameter is true. However, the SDK `Ip::index()` method always defaults to passing false, and has no ability to return all ips.

This PR fixes that so the default behavior for `Ip::index()` is to return all IPs (instead of shared IPs). This is more consistent with the API.

### DIFF
--- a/tests/Api/Ip.php
+++ b/tests/Api/Ip.php
@@ -1,0 +1,85 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * Copyright (C) 2013 Mailgun
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+namespace Mailgun\Tests\Api;
+
+use GuzzleHttp\Psr7\Response;
+use Mailgun\Api\Ip;
+use Mailgun\Model\Ip\IndexResponse;
+
+class IpTest extends TestCase
+{
+    protected function getApiClass()
+    {
+        return Ip::class;
+    }
+
+    public function testIndexAll()
+    {
+        $this->setRequestMethod('GET');
+        $this->setRequestUri('/v3/ips');
+        $this->setHttpResponse(new Response(200, ['Content-Type' => 'application/json'], <<<'JSON'
+{
+  "items": ["192.161.0.1", "192.168.0.2"],
+  "total_count": 2
+}
+JSON
+        ));
+
+        $api = $this->getApiInstance();
+        /** @var IndexResponse $response */
+        $response = $api->index(null);
+        $this->assertInstanceOf(IndexResponse::class, $response);
+        $this->assertEquals(2, $response->getTotalCount());
+        $this->assertEquals('192.161.0.1', $response->getItems()[0]);
+        $this->assertEquals('192.168.0.2', $response->getItems()[1]);
+    }
+    
+    public function testIndexOnlyDedicated()
+    {
+        $this->setRequestMethod('GET');
+        $this->setRequestUri('/v3/ips?dedicated=1');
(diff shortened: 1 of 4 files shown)


tests/Api/Ip.php
```diff
@@ -0,0 +1,85 @@
...
+        $this->assertInstanceOf(IndexResponse::class, $response);
+        $this->assertEquals(2, $response->getTotalCount());
+        $this->assertEquals('192.161.0.1', $response->getItems()[0]);
+        $this->assertEquals('192.168.0.2', $response->getItems()[1]);
+    }
+    
+    public function testIndexOnlyDedicated()
+    {
+        $this->setRequestMethod('GET');
+        $this->setRequestUri('/v3/ips?dedicated=1');
```
DavidGarciaCat, line 49: I am not sure about this one. According to the docs, we are supposed to pass `true` instead of `1`

https://documentation.mailgun.com/en/latest/api-ips.html#ips

image

Moreover, as part of the example, we are supposed to pass that `true` value as a string instead of an integer or boolean (at least, as defined for the `curl` call):

https://documentation.mailgun.com/en/latest/api-ips.html#example

image

Are we 100% sure this is the case? I would prefer not merging this until we can confirm it won't break the implementation.

DavidGarciaCat (commented):
Hello @josephshanak 

Thanks for your PR.

I just have a question, that it brings me a bit of doubt in regards to this update, compared with the API Docs on the official Mailguin website.

Can you please run a curl command and paste a screenshot with the result so we can get confirmation that it works?

Thanks,

tests/Api/Ip.php
josephshanak: I ran the following commands (I edited the results to remove the API key):

```
$ curl -s --user 'api:API_KEY' -G https://api.mailgun.net/v3/ips?dedicated=false
{"assignable_to_pools":["159.135.238.30"],"items":["209.61.151.242"],"total_count":1}

$ curl -s --user 'api:API_KEY' -G https://api.mailgun.net/v3/ips?dedicated=true
{"assignable_to_pools":["159.135.238.30"],"items":["159.135.225.45","166.78.69.72","159.135.238.30"],"total_count":3}

$ curl -s --user 'api:API_KEY' -G https://api.mailgun.net/v3/ips
{"assignable_to_pools":["159.135.238.30"],"items":["159.135.225.45","166.78.69.72","159.135.238.30","209.61.151.242"],"total_count":4}

$ curl -s --user 'api:API_KEY' -G https://api.mailgun.net/v3/ips?dedicated=0
{"assignable_to_pools":["159.135.238.30"],"items":["209.61.151.242"],"total_count":1}

$ curl -s --user 'api:API_KEY' -G https://api.mailgun.net/v3/ips?dedicated=1
{"assignable_to_pools":["159.135.238.30"],"items":["159.135.238.30","159.135.225.45","166.78.69.72"],"total_count":3}
```

It appears that:
1. `0`/`1` return the same results as `false`/`true` for the dedicated parameter
2. Without specifying the parameter `dedicated` returns different results than either `true`/`false`/`0`/`1`.
3. There is an additional key in the response `assignable_to_pools` which doesn't appear in the documentation

DavidGarciaCat: :+1: for points 1 and 2

Maybe I would suggest, If you don't mind, adding the mapping for the property that we don't have at the current stage, so that we can cover this case too?

josephshanak: I added `assignable_to_pools` and pushed a change. See commit 8c092ab5952b82a2207cc251345483082d3ded4f

DavidGarciaCat (approved):
Much appreciated

PR #709, opened 2020-10-08
Update Event.php

dannoso:
Fix broken pagination when invoking events()->get()

### DIFF
--- a/src/Api/Event.php
+++ b/src/Api/Event.php
@@ -33,12 +33,12 @@ class Event extends HttpApi
         if (array_key_exists('limit', $params)) {
             Assert::range($params['limit'], 1, 300);
         }
-        
+
         if ($page) {
             $page = explode('/events', $page)[1];
         }
 
-        $response = $this->httpGet(sprintf('/v3/%s/events' . $page, $domain), $params);
+        $response = $this->httpGet(sprintf('/v3/%s/events'.$page, $domain), $params);
 
         return $this->hydrateResponse($response, EventResponse::class);
     }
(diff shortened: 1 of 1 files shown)


DavidGarciaCat (commented):
Hello,

Can you please justify the decision behind this update?

The `API\Event` class is returning an `EventResponse` object, as you can see either on the `@return` annotation as well as on the `return` call inside the method:

https://github.com/mailgun/mailgun-php/blob/27d579fdf1295dd3fdf35f14cde797de21615fb7/src/Api/Event.php#L26-L40

And the `EventResponse` class is implementing the required code to deal with the pagination, using the expected Trait for that purpose:

https://github.com/mailgun/mailgun-php/blob/27d579fdf1295dd3fdf35f14cde797de21615fb7/src/Model/Event/EventResponse.php#L21-L23

It means that, once you get the response, you can call the already included methods for browsing either the next or previous page:

https://github.com/mailgun/mailgun-php/blob/27d579fdf1295dd3fdf35f14cde797de21615fb7/src/Model/PaginationResponse.php#L24-L32

Please let me know your point, as I don't see any need for this PR.

Thanks,

DavidGarciaCat:
Closing this PR for now, as it seems to be unnecessary due to we already implement the required code to work with the pagination.

If someone can prove this is really a requirement, we can discuss a new implementation, including the expected unit tests and following the same schema that we already have as part of any other call.

dannoso:
Hi David, thank for the review.
I realized that right after I opened the PR but I keep it anyway because I think that be able to use the same method you use in the first call also for the pagination makes a lot of sense and it is expected.
As for now you have to make an unnecessary exception to call the next pages: another method with another signature to basically get the next chuck of same result.

I know this might redundant but it gives another option to people like me :)

Thanks,
Davide

Issue #633, opened 2019-09-13
New implementation for deprecated or new functionality


DavidGarciaCat:
While updating the Mailgun API Docs and as per the issues that have been reported so far, the following functionalities are working based on the deprecated endpoints. New Pull Requests are accepted to expedite the new functionalities.

### ~Domains Tracking~

~https://documentation.mailgun.com/en/latest/api-domains.html?highlight=tracking#domains~

### ~Email Validation~

~https://documentation.mailgun.com/en/latest/api-email-validation.html~

~The API v4 for email validation requires a new implementation.~

~Please refer to the issues https://github.com/mailgun/mailgun-php/issues/617 and https://github.com/mailgun/mailgun-php/issues/619 and to the Pull Request https://github.com/mailgun/mailgun-php/issues/618 for additional information.~

### Mailboxes

https://documentation.mailgun.com/en/latest/api-mailboxes.html

### ~Mailing Lists validation~

~https://documentation.mailgun.com/en/latest/api-mailinglists.html~

### ~Webhooks~

~https://documentation.mailgun.com/en/latest/api-webhooks.html~

~The IDs have been renamed, and we can now pass an array of URLs instead of a single one.~

~Please refer to the Pull Request https://github.com/mailgun/mailgun-php/pull/609 and https://github.com/mailgun/mailgun-php/pull/749 for additional information.~

### ~Whitelist Suppression~

~https://documentation.mailgun.com/en/latest/api-suppressions.html#whitelists~

uavn:
@DavidGarciaCat 

Since Domain Tracking, Mailing List Validation and Suppression Whitelists are done, 
Webhooks has already PR creted,
Mailboxes are deprecated (see the link: Warning Mailboxes is a legacy feature of Mailgun that is no longer offered to new customers. Please utilize the Store() endpoint instead.)

what else should be done here?

DavidGarciaCat:
Hello @uavn 

We should review the [API Documentation](https://documentation.mailgun.com/en/latest/api_reference.html) and confirm if there's any other missing functionality. In the best-case scenario, everything will already be implemented.

And, although I have been saying it every time you pushed a PR, thank you very much for all your contributions.

instalab:
Hi
[...]

PR #672, opened 2020-03-17
Implements domain tracking API.

BlakeHancock:
Implements the following API endpoints.

GET /domains/<domain>/tracking
PUT /domains/<domain>/tracking/open
PUT /domains/<domain>/tracking/click
PUT /domains/<domain>/tracking/unsubscribe

PR #711, opened 2020-10-09
Domain tracking

uavn:
#633

### DIFF
--- a/src/Api/Domain.php
+++ b/src/Api/Domain.php
@@ -325,8 +325,8 @@ class Domain extends HttpApi
      *
      * @param string $domain name of the domain
      *
-     * @param string $active
      * @return UpdateClickTrackingResponse|array|ResponseInterface
+     *
      * @throws \Exception
      */
     public function updateClickTracking(string $domain, string $active)
@@ -348,7 +348,7 @@ class Domain extends HttpApi
      * Updates a domain open tracking settings.
      *
      * @param string $domain name of the domain
-     * @param bool $active
+     * @param bool   $active
      *
      * @return UpdateOpenTrackingResponse|array|ResponseInterface
      */
@@ -370,10 +370,9 @@ class Domain extends HttpApi
     /**
      * Updates a domain unsubscribe tracking settings.
      *
-     * @param string $domain name of the domain
-     * @param string $active
-     * @param string $htmlFooter
+     * @param string      $domain     name of the domain
      * @param string|null $textFooter
+     *
      * @return UpdateUnsubscribeTrackingResponse|array|ResponseInterface
      */
     public function updateUnsubscribeTracking(string $domain, string $active, string $htmlFooter, string $textFooter)
@@ -385,7 +384,7 @@ class Domain extends HttpApi
         Assert::nullOrString($textFooter);
 
         $params = [
-            'active' => (in_array($active, ['yes', 'true'])) ? 'true' : 'false',
+            'active' => (in_array($active, ['yes', 'true'], true)) ? 'true' : 'false',
             'html_footer' => $htmlFooter,
             'text_footer' => $textFooter,
         ];
--- a/src/Api/Domain.php
+++ b/src/Api/Domain.php
@@ -325,13 +325,18 @@ class Domain extends HttpApi
      *
      * @param string $domain name of the domain
      *
-     * @param bool $active
+     * @param string $active
      * @return UpdateClickTrackingResponse|array|ResponseInterface
+     * @throws \Exception
      */
-    public function updateClickTracking(string $domain, bool $active)
+    public function updateClickTracking(string $domain, string $active)
     {
+        Assert::stringNotEmpty($domain);
+        Assert::stringNotEmpty($active);
+        Assert::oneOf($active, ['yes', 'no', 'htmlonly']);
+
         $params = [
-            'active' => $active ? 'true' : 'false',
+            'active' => $active,
         ];
 
         $response = $this->httpPut(sprintf('/v3/domains/%s/tracking/click', $domain), $params);
@@ -347,10 +352,14 @@ class Domain extends HttpApi
      *
      * @return UpdateOpenTrackingResponse|array|ResponseInterface
      */
-    public function updateOpenTracking(string $domain, bool $active)
+    public function updateOpenTracking(string $domain, string $active)
     {
+        Assert::stringNotEmpty($domain);
+        Assert::stringNotEmpty($active);
+        Assert::oneOf($active, ['yes', 'no']);
+
         $params = [
-            'active' => $active ? 'true' : 'false',
+            'active' => $active,
         ];
 
         $response = $this->httpPut(sprintf('/v3/domains/%s/tracking/open', $domain), $params);
@@ -362,15 +371,21 @@ class Domain extends HttpApi
      * Updates a domain unsubscribe tracking settings.
      *
      * @param string $domain name of the domain
-     * @param bool $active
+     * @param string $active
      * @param string $htmlFooter
-     * @param string $textFooter
+     * @param string|null $textFooter
      * @return UpdateUnsubscribeTrackingResponse|array|ResponseInterface
      */
-    public function updateUnsubscribeTracking(string $domain, bool $active, string $htmlFooter, string $textFooter)
+    public function updateUnsubscribeTracking(string $domain, string $active, string $htmlFooter, string $textFooter)
     {
+        Assert::stringNotEmpty($domain);
+        Assert::stringNotEmpty($active);
+        Assert::oneOf($active, ['yes', 'no', 'true', 'false']);
+        Assert::stringNotEmpty($htmlFooter);
+        Assert::nullOrString($textFooter);
+
         $params = [
-            'active' => $active ? 'true' : 'false',
+            'active' => (in_array($active, ['yes', 'true'])) ? 'true' : 'false',
             'html_footer' => $htmlFooter,
             'text_footer' => $textFooter,
         ];
--- a/src/Api/Domain.php
+++ b/src/Api/Domain.php
@@ -20,8 +20,12 @@ use Mailgun\Model\Domain\DeleteCredentialResponse;
 use Mailgun\Model\Domain\DeleteResponse;
 use Mailgun\Model\Domain\IndexResponse;
 use Mailgun\Model\Domain\ShowResponse;
+use Mailgun\Model\Domain\TrackingResponse;
+use Mailgun\Model\Domain\UpdateClickTrackingResponse;
 use Mailgun\Model\Domain\UpdateConnectionResponse;
 use Mailgun\Model\Domain\UpdateCredentialResponse;
+use Mailgun\Model\Domain\UpdateOpenTrackingResponse;
+use Mailgun\Model\Domain\UpdateUnsubscribeTrackingResponse;
 use Mailgun\Model\Domain\VerifyResponse;
 use Psr\Http\Message\ResponseInterface;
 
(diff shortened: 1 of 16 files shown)


src/Api/Domain.php
```diff
@@ -299,4 +303,80 @@ public function verify(string $domain)
...
+     *
+     * @param string $domain name of the domain
+     *
+     * @param bool $active
+     * @return UpdateClickTrackingResponse|array|ResponseInterface
+     */
+    public function updateClickTracking(string $domain, bool $active)
+    {
+        $params = [
+            'active' => $active ? 'true' : 'false',
```
DavidGarciaCat, line 334: According to the Mailgun Documentation, the value should be `yes`, `no` or `htmlonly` instead of `true` or `false`.

Can you fix this implementation, please, and make sure the automated tests are checking the expected values?

Please don't forget adding the expected `Asset::...` before setting the parameters. It would be great if you also add additional unit tests to validate the exceptions thrown by the `Asset` methods.

https://documentation.mailgun.com/en/latest/api-domains.html#domains

image

src/Api/Domain.php
```diff
@@ -299,4 +303,80 @@ public function verify(string $domain)
...
+     *
+     * @param string $domain name of the domain
+     * @param bool $active
+     *
+     * @return UpdateOpenTrackingResponse|array|ResponseInterface
+     */
+    public function updateOpenTracking(string $domain, bool $active)
+    {
+        $params = [
+            'active' => $active ? 'true' : 'false',
```
DavidGarciaCat, line 353: According to the Mailgun Documentation, the value should be `yes` or `no` instead of `true` or `false`.

Can you fix this implementation, please, and make sure the automated tests are checking the expected values?

Please don't forget adding the expected `Asset::...` before setting the parameters. It would be great if you also add additional unit tests to validate the exceptions thrown by the `Asset` methods.

https://documentation.mailgun.com/en/latest/api-domains.html#domains

image

src/Api/Domain.php
```diff
@@ -299,4 +303,80 @@ public function verify(string $domain)
...
+     * @param string $htmlFooter
+     * @param string $textFooter
+     * @return UpdateUnsubscribeTrackingResponse|array|ResponseInterface
+     */
+    public function updateUnsubscribeTracking(string $domain, bool $active, string $htmlFooter, string $textFooter)
+    {
+        $params = [
+            'active' => $active ? 'true' : 'false',
+            'html_footer' => $htmlFooter,
+            'text_footer' => $textFooter,
```
DavidGarciaCat, line 375: According to the Mailgun Documentation, it's not clear for me if the values for the `active` property must be a `bool` or a `string`. However, it seems like at least the `text_footer` might be optional, given the API documentation states that it can be generated automatically by the Mailgun system.

Can you fix this implementation, please, making sure we are adding the parameters just if they are strictly mandatory, and not adding them when they are optional?

Please don't forget adding the expected `Asset::...` before setting the parameters. It would be great if you also add additional unit tests to validate the exceptions thrown by the `Asset` methods.

https://documentation.mailgun.com/en/latest/api-domains.html#domains

image

src/Model/Domain/TrackingResponse.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+        if (!isset($data['tracking'])) {
+            return null;
+        }
+
+        $trackingSettings = $data['tracking'];
+
+        $model = new self();
+        $model->click = ClickTracking::create($trackingSettings['click'] ?? []);
+        $model->open = OpenTracking::create($trackingSettings['click'] ?? []);
```
DavidGarciaCat, line 35: It seems to me that this property should be checking the `open` key of the array instead of the `click` one. Can you please make sure this is fixed?

Also, as part of the automated tests, when simulating the response, please try to set one of the properties as `true` and the other one as `false` so you can confirm they are not mapped wrongly when copying and pasting.

Please note the last comment is a suggestion for future cases too, so you can confirm the tests are providing more accurate results, given is hard getting rid of a bad habit such as copy-paste.

DavidGarciaCat (changes requested):
Hello, @uavn 

Firstly, thanks for sending this PR.

I have seen a few lines that are a bit concerning to me, so I have added inline comments for each one of them, making it easier to find and apply the updates.

If you don't mind taking a look and sending another commit with the amends, I am more than happy to take a look to the updated PR and merge the new code into the master branch, so we can release a new version soon with your changes.

As a minor comment, please don't forget adding the `Asset::...` methods for future updates, as we need to make sure the provided data makes sense and matches with the expected content, so we can prevent unexpected errors. Thanks for your understanding.

Thanks for contributing to this repo.

Cheers,

DavidGarciaCat:
By the way: do you mind taking a look to the failures reported by `PrettyCI` too, please? Would be good if we get the thumbs up from that one too, so we keep a better formatted and cleaner code. Thanks,

src/Api/Domain.php
uavn: All done, except the optional text_footer, because seems like its mandatory, because API returns an error:
```
{
  "message": "Missing parameter 'text_footer'"
}
```
when I'm not passing text_footer or passing it empty

DavidGarciaCat:
Thanks for the updates, @uavn - if you don't mind, I would love to test it first before merging. But knowing the code is supposed to be ready, I will try to do it as soon as possible.

uavn:
> Thanks for the updates, @uavn - if you don't mind, I would love to test it first before merging. But knowing the code is supposed to be ready, I will try to do it as soon as possible.

sure thing

src/Api/Domain.php
```diff
@@ -299,4 +303,94 @@ public function verify(string $domain)
...
+
+        $response = $this->httpGet(sprintf('/v3/domains/%s/tracking', $domain));
+
+        return $this->hydrateResponse($response, TrackingResponse::class);
+    }
+
+    /**
+     * Updates a domain click tracking settings.
+     *
+     * @param string $domain name of the domain
```
DavidGarciaCat, line 326: Missing annotation for the 2nd argument
```suggestion
     * @param string $domain The name of the domain
     * @param string $active The status for this tracking (one of: yes, no)
```

src/Api/Domain.php
```diff
@@ -299,4 +303,94 @@ public function verify(string $domain)
...
+
+        $response = $this->httpPut(sprintf('/v3/domains/%s/tracking/click', $domain), $params);
+
+        return $this->hydrateResponse($response, UpdateClickTrackingResponse::class);
+    }
+
+    /**
+     * Updates a domain open tracking settings.
+     *
+     * @param string $domain name of the domain
```
DavidGarciaCat, line 350: Missing annotation for the 2nd argument

```suggestion
     * @param string $domain The name of the domain
     * @param string $active The status for this tracking (one of: yes, no)
```

src/Api/Domain.php
```diff
@@ -299,4 +303,94 @@ public function verify(string $domain)
...
+
+        $response = $this->httpPut(sprintf('/v3/domains/%s/tracking/open', $domain), $params);
+
+        return $this->hydrateResponse($response, UpdateOpenTrackingResponse::class);
+    }
+
+    /**
+     * Updates a domain unsubscribe tracking settings.
+     *
+     * @param string $domain name of the domain
```
DavidGarciaCat, line 372: Missing annotations for the arguments

```suggestion
     * @param string $domain     The name of the domain
     * @param string $active     The status for this tracking (one of: yes, no)
     * @param string $htmlFooter The footer for HTML emails
     * @param string $textFooter The footer for plain text emails
```

src/Model/Domain/ClickTracking.php
```diff
@@ -0,0 +1,57 @@
...
+        $model = new self();
+        $active = $data['active'] ?? null;
+
+        if (true === $active) {
+            $model->active = 'yes';
+        } elseif (false === $active) {
+            $model->active = 'no';
+        } else {
+            $model->active = $active;
+        }
```
DavidGarciaCat, line 34: Suggesting a ternary operator instead of concatenating a set of conditionals, as it makes the code shorter

```php
        $active = $data['active'] ?? null;

        $model = new self();
        $model->active = 'htmlonly' === $active ? $active : ($active ? 'yes' : 'no');
```

src/Model/Domain/OpenTracking.php
```diff
@@ -0,0 +1,55 @@
...
+        $model = new self();
+        $active = $data['active'] ?? null;
+
+        if (true === $active) {
+            $model->active = 'yes';
+        } elseif (false === $active) {
+            $model->active = 'no';
+        } else {
+            $model->active = $active;
+        }
```
DavidGarciaCat, line 34: Three cases seems unnecessary, given this case doesn't allow the `htmlonly` according to the API Docs

```php
        $model = new self();
        $model->active = $active ? 'yes' : 'no';
```

src/Model/Domain/UnsubscribeTracking.php
```diff
@@ -0,0 +1,58 @@
...
+final class UnsubscribeTracking
+{
+    private $active;
+    private $htmlFooter;
+    private $textFooter;
+
+    public static function create(array $data): self
+    {
+        $model = new self();
+        $model->active = ($data['active'] ?? null) ? 'true' : 'false';
```
DavidGarciaCat, line 28: Suggesting setting `yes` or `no` so that we can keep consistency with the other two models

```suggestion
        $model->active = ($data['active'] ?? null) ? 'yes' : 'no';
```

src/Model/Domain/UnsubscribeTracking.php
```diff
@@ -0,0 +1,58 @@
...
+        return $model;
+    }
+
+    private function __construct()
+    {
+    }
+
+    public function isActive(): bool
+    {
+        return 'true' === $this->getActive();
```
DavidGarciaCat, line 41: Due to the previous comment, this one needs to be updated too

```suggestion
        return 'yes' === $this->getActive();
```

tests/Model/Domain/TrackingResponseTest.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        $json =
+            <<<'JSON'
+{
+    "tracking": {
+        "click": {
+            "active": "htmlonly"
+        },
+        "open": {
+            "active": "no"
```
DavidGarciaCat, line 32: According to the example at the end of the page of the API docs, this one seems to be a boolean value

```suggestion
            "active": false
```

https://documentation.mailgun.com/en/latest/api-domains.html#example

image

tests/Model/Domain/TrackingResponseTest.php
```diff
@@ -0,0 +1,61 @@
...
+        $this->assertTrue($model->getClick()->isHtmlOnly());
+
+        $this->assertNotEmpty($model->getOpen());
+        $this->assertInstanceOf(OpenTracking::class, $model->getOpen());
+        $this->assertEquals('no', $model->getOpen()->getActive());
+        $this->assertFalse($model->getOpen()->isActive());
+
+        $this->assertNotEmpty($model->getUnsubscribe());
+        $this->assertInstanceOf(UnsubscribeTracking::class, $model->getUnsubscribe());
+        $this->assertEquals('false', $model->getUnsubscribe()->getActive());
```
DavidGarciaCat, line 56: Due to the previous change on the source code, aiming to keep consistency, it needs to be updated too

```suggestion
        $this->assertEquals('no', $model->getUnsubscribe()->getActive());
```

tests/Model/Domain/UpdateOpenTrackingResponseTest.php
```diff
@@ -0,0 +1,38 @@
...
+
+class UpdateOpenTrackingResponseTest extends BaseModelTest
+{
+    public function testCreate()
+    {
+        $json =
+            <<<'JSON'
+{
+  "open": {
+    "active": "no"
```
DavidGarciaCat, line 26: As mentioned in the other file, it seems to be a boolean as part of the API response

```suggestion
    "active": false
```

DavidGarciaCat (commented):
Hello @uavn 

I am adding a few suggestions (some of them can be applied immediately through GitHub itself) due to two reasons:
- On some cases, we don't keep consistency with our models and responses. I would prefer to keep that consistency across our models, so we have a standard format.
- There are a couple of cases where we don't keep consistency with the output returned (according to the API docs)

I also tested the code locally with one of my domains, and all these comments should be working without any error, so if you have 5 minutes to apply the suggestions, we can merge and tag a new release.

Many thanks for your time contributing to this feature.

Cheers,

DavidGarciaCat:
Reminder: merging this PR means we can close this other PR (https://github.com/mailgun/mailgun-php/pull/672) sent some time ago but (apparently) inactive.

uavn:
all comments fixed

DavidGarciaCat:
Thanks

PR #714, opened 2020-10-16
Mailing list; Create multiple members subscribed fix

thecaliskan:
Hi; 

Add a mailing list multiple members function is broken;

Example

Create function 

> 
> PUT /lists/<address\>/members/<member_address>
> 
> 'subscribed' parameter is string (yes or no)

but Create multiple function 

> POST /lists/<address\>/members.json
> 
> 'subscribed' parameter is boolean (yes or no)

You can check the following url

[https://documentation.mailgun.com/en/latest/api-mailinglists.html#mailing-lists](https://documentation.mailgun.com/en/latest/api-mailinglists.html#mailing-lists)


SonerOd:
> Hi;
> 
> Add a mailing list multiple members function is broken;
> 
> Example
> 
> Create function
> 
> > PUT /lists/<address>/members/<member_address>
> > 'subscribed' parameter is string (yes or no)
> 
> but Create multiple function
> 
> > POST /lists/<address>/members.json
> > 'subscribed' parameter is boolean (yes or no)
> 
> You can check the following url
> 
> https://documentation.mailgun.com/en/latest/api-mailinglists.html#mailing-lists

+1

DavidGarciaCat (changes requested):
Hello @thecaliskan 

Thanks for your PR

Would you mind please updating the unit tests according to the updated validation that you have updated? There is now an inconsistency that Travis CI reports, and it's impossible to merge it (not to mention that could make the system unstable and confusing for other developers working on other updates).

Thank you,

thecaliskan:
Hello,

Updated unit test.

Thank you

thecaliskan:
Hello again;

Added missing stats of event types, thanks



[https://documentation.mailgun.com/en/latest/api-stats.html#event-types](https://documentation.mailgun.com/en/latest/api-stats.html#event-types)

thecaliskan:
Hello @DavidGarciaCat 

Review, Answer ?

DavidGarciaCat:
Hello @thecaliskan 

Thank you for sending this PR, as well as for the more recent commits.

I have two main points that I want to share with you, and I really hope they both make sense:

----

**Inconsistent PR**

I can see that this PR is no longer focused on solving `one` "potential" (or not) issue that is related to consistency (between `strings` and `booleans`). Now it tries to cover several things at once, given there are also changes in `Stats` and `Tags`.

I don't think I should merge this PR as it has lost its focus/meaning. I would definitively prefer getting several smaller PRs for better code reviews and ensure we're not mixing different concepts, so each one of them will keep their focus on just one fix at a time.

----

**Potential SemVer BC break**

As you might be aware of, semantic versioning states software versions as `MAJOR.MINOR.PATCH`. Changing the internal logic of the SDK to operate completely with `boolean` instead of `string` means we will get a new BC break unless we release a new major version.

Given this change is not adding any significant change but a fix for an inconsistent value, I would prefer getting an update where the input data accepts either a `bool` or a `string`. In this way, we won't cause any BC break while the developer will get more freedom to choose how to progress. When the time comes, we can decide which one should be deprecated and which one should remain.

Having said that, if you want to work on this change, then we should prioritise the `string` over the `boolean` value as there are some cases where we can pass three values instead of two (such as `yes`/`no`/`htmlonly`) forcing a `bool` might lead to a misinterpretation of the options available.

----

Please let me know if these comments make sense to you. As mentioned, I am more than happy to deal with the fixes, as far as we keep consistency and focus on what we try to fix by creating different branches.

Thanks in advance,

D

Nyholm:
I vote for closing this PR or that you address [the comments](https://github.com/mailgun/mailgun-php/pull/714#issuecomment-714632069) from David.

DavidGarciaCat:
Probably will be harder to amend this PR instead of creating two or three new updates?

DavidGarciaCat:
Closing this PR as per Nyholm's suggestion here https://github.com/mailgun/mailgun-php/pull/714#issuecomment-770168334 and also because of the mix of changes reported here https://github.com/mailgun/mailgun-php/pull/714#issuecomment-714632069

We can create a new set of updates to cover the different changes, assuming they are still a problem.

PR #721, opened 2020-11-16
Add support for PHP 8.x

DavidGarciaCat:


DavidGarciaCat:
Just the most recent PHP Unit versions allow installing it with PHP versions >= 7.3

Example: https://packagist.org/packages/phpunit/phpunit#9.4.3

However, these versions are conflicting with previous versions of the SDK as the support for PHP 7.x (including 7.1 and 7.2) will break because now PHP Unit requires setting `: void` for the `setup` and `teardown` functions (among others).

Upgrade the SDK to be used with PHP 8.0 will cause a BC Break, meaning that it cannot be considered at the current stage.

PR #722, opened 2020-11-16
Prepare for Mailgun SDK v3.2.0

DavidGarciaCat:


Nyholm:
Nice. Thank you. 

There are a few more things I would like to add. But that can be done later :)

Issue #619, opened 2019-08-11
Review the Email Validation functionality


DavidGarciaCat:
As pointed out here #617 the validation didn't work when tried it using the US servers.

The Mailgun Support team has confirmed that the service should work as follows:

---

> I can confirm the EU region does not have email validation at this time; only the US.  As for v3 of the email validation API, it is still operative.  I just ran a quick test just to confirm:

```
$ curl -G --user 'api:pubkey-REDACTED' -G \
>    https://api.mailgun.net/v3/address/validate \
>    --data-urlencode address='mgbox01@gmail.com'
```
```
{
    "address":"mgbox01@gmail.com",
    "did_you_mean":null,
    "is_disposable_address":false,
    "is_role_address":false,
    "is_valid":true,
    "mailbox_verification":"true",
    "parts":{
        "display_name":null,
        "domain":"gmail.com",
        "local_part":"mgbox01"
    },
    "reason":null
}
```

> And we moved the documentation for it to the following URL: https://documentation.mailgun.com/en/latest/api-email-validation-deprecated.html#email-validation

---

Thus, the SDK needs an update to cover the following cases:
- Potential check and error message if someone tries to validate an email through the EU service
- Potential check and error message if someone tries to validate an email through a private key
- Potential adjustments to use the API v4 instead of the API v3 for the email validation

Even if we don't update the code, then we should consider to add a disclaimer as part of the README file to announce these details and prevent other misunderstandings.

saturnonearth:
Hi there,

Yes I have confirmed with support as well that 

> They are using v3 vs the v4 Email Validation API. the v3 is is_valid just means it’s a RFC syntactical address, it doesn’t perform mailbox verification unless specified. We recommend them use the v4 email validation API

Here's my vote for V4 implementation 🙌😎

pinepain:
Hi, any progress on this issue? Is help needed?

natatkinson:
I'm working with mailgun support and they said I should be using v4 of validation. Any idea on when this might be ready?

Matroskeen:
Hello guys,
Any news so far? Do you have any estimates of the completion?

We would like to integrate Email Validation into Drupal module and I think it'll be great to use v4 version: https://www.drupal.org/project/mailgun/issues/2966690#comment-13400874

saturnonearth:
> Hello guys,
> Any news so far? Do you have any estimates of the completion?
> 
[...]

PR #726, opened 2020-11-25
email validation v4, bulk validation, bulk preview

uavn:
https://github.com/mailgun/mailgun-php/issues/619


DavidGarciaCat:
Funny thing you've sent this update, as I was thinking about how to implement this feature with the fewer changes possible for the last couple of days.

I can see it is still flagged as a Draft, so please let me know once it's ready, and I'll start the review/comments (I already saw a couple of things that I would prefer not to have them in the main branch).

And thank you for working on this update, as it is highly requested.

uavn:
> Funny thing you've sent this update, as I was thinking about how to implement this feature with the fewer changes possible for the last couple of days.
> 
> I can see it is still flagged as a Draft, so please let me know once it's ready, and I'll start the review/comments (I already saw a couple of things that I would prefer not to have them in the main branch).
> 
> And thank you for working on this update, as it is highly requested.

Hi, yes, I want to make a few more tweaks and test it a bit more, then will move PR from draft state

uavn:
@DavidGarciaCat tested, added a few fixes, PR is ready for review

kslr:
any update?

uavn:
> any update?
@DavidGarciaCat

lptn:
any update?
@DavidGarciaCat

kslr:
Mailgun takes several months each time, I have given up using it

DavidGarciaCat:
I will try to invest some time this month. Sorry for the delay, guys, but I'm a bit busy with several open fronts and I need to address all of them in the best possible way.

DavidGarciaCat:
@kslr Please note this is an open-source repo running under the MIT license and maintained by self-volunteers. You can expedite the development and review phases by being part of the community. Please don't bring any unnecessary arguments here, as we are not getting paid to keep the SDK updated.

src/Api/EmailValidationV4.php
```diff
@@ -0,0 +1,282 @@
...
+            // File from memory
+            $resource = fopen('php://temp', 'r+');
+            fwrite($resource, $filePath['fileContent']);
+            rewind($resource);
+        } elseif (isset($filePath['filePath'])) {
+            // File form path
+            $path = $filePath['filePath'];
+
+            // Remove leading @ symbol
+            if (0 === strpos($path, '@')) {
```
DavidGarciaCat, line 253: What will happen if we have 2x `@`? Maybe is safer running a regular expression with `preg_replace`?

image

src/Api/EmailValidationV4.php
```diff
@@ -0,0 +1,282 @@
...
+                $path = substr($path, 1);
+            }
+
+            $resource = fopen($path, 'r');
+        } else {
+            throw new InvalidArgumentException('When using a file you need to specify parameter "fileContent" or "filePath"');
```
DavidGarciaCat, line 259: I believe it's safe to be deleted.

Given this is an internal (private) method, just accessible by this API class, it's my understanding this `throw` becomes an unreachable statement, as the other two conditionals are checking for the two allowed options, and the user doesn't get a chance to set them dynamically when calling the method.

src/Api/HttpApi.php
```diff
@@ -95,6 +95,8 @@ protected function handleErrors(ResponseInterface $response)
                 throw  HttpClientException::forbidden($response);
             case 404:
                 throw HttpClientException::notFound($response);
+            case 409:
+                throw HttpClientException::conflict($response);
```
DavidGarciaCat, line 99: Did you see this error happening? I can't see any reference to the `409 Conflict` status code on the official documentation, and I don't remember seeing it while I've been working with this part APIs for a long time.

In case it's unseen, then I believe doesn't make much sense including it, meaning that we can also undo the update here: `src/Exception/HttpClientException.php`

DavidGarciaCat (commented):
Hi @uavn 

First things first, thank you for working and sending this PR. Indeed, it is an important contribution to the SDK, and I am eager to see it working.

Please accept my apologies for my delay on this review. I know it has been here for a while, but some times life goes in the way, and I needed (and still need) to attend other matters.

I have added a couple of minor points that you can easily deal with them. After pushing them, I think this PR is ready to be merged and tagged.

Once more, thank you very much for your contribution.

D

nicodemuz:
Tested out the `emailValidationV4`, seems to work great! :+1:

Would be great if there were some constants for some of the response fields, so I could avoid comparing to the string itself, e.g.:
```
        if ($this->mailgun->emailValidationV4()->validate($emailAddress)->getResult() === 'deliverable') {
            return;
        }
```

DavidGarciaCat:
Setting the constants might be something you define in your app, as it will match with your logic and expectations. The SDK provides the functionality to run the operations you need.

I am more concerned about not hearing from @uavn although some times life goes in the way.

uavn:
@DavidGarciaCat pardon, I was just busy on the holidays, I will fix those things in nearest days

src/Api/EmailValidationV4.php
uavn: @DavidGarciaCat So this is needed when you have a file [file.txt], try to set filepath as:
'@file.txt' like here: https://github.com/mailgun/mailgun-php/blob/master/tests/Message/MessageBuilderTest.php#L146

and this condition will remove the leading zero, the same condition exists here: https://github.com/mailgun/mailgun-php/blob/master/src/Api/Message.php#L141

src/Api/EmailValidationV4.php
uavn: @DavidGarciaCat removed

src/Api/HttpApi.php
uavn: @DavidGarciaCat Yes, it happens:
```
try {
    $listId = 'testListId2';

    $response = $mg->emailValidationV4()
        ->deleteBulkJob($listId);
} catch (Exception $e) {
}
```
```
public function deleteBulkJob(string $listId)
{
    Assert::stringNotEmpty($listId);

    $response = $this->httpDelete(sprintf('/v4/address/validate/bulk/%s', $listId));

    print_r($response);

    return $this->hydrateResponse($response, DeleteBulkJobResponse::class);
}
```
```
root@c33e913ec8cd:/var/www/mailgun-php-client# php index.php 
GuzzleHttp\Psr7\Response Object
(
    [reasonPhrase:GuzzleHttp\Psr7\Response:private] => Conflict
    [statusCode:GuzzleHttp\Psr7\Response:private] => 409
    [headers:GuzzleHttp\Psr7\Response:private] => Array
...
```

src/Api/Message.php
```diff
@@ -144,8 +144,6 @@ private function prepareFile(string $fieldName, array $filePath): array
             }
 
             $resource = fopen($path, 'r');
-        } else {
-            throw new InvalidArgumentException('When using a file you need to specify parameter "fileContent" or "filePath"');
```
uavn: removed in wrong place

uavn:
@DavidGarciaCat done

src/Api/EmailValidationV4.php
Nyholm: I think this should be removed. It is a legacy layer for new code.

If we fear that some will use a file path like `@/foo/bar.txt` then we should update the documentation.

Nyholm (commented):
Thank you. 
Im happy with this.

DavidGarciaCat:
After the final bit to be amended, it should be rebased to avoid conflicts, as it cannot be merged now

image

uavn:
@Nyholm @DavidGarciaCat 
condition removed, branch rebased

DavidGarciaCat (approved):
Thank you for your contribution. This is, indeed, a very expected update. :-)

Nyholm:
Wohoo. 

Thank you

RayHughes:
Seriously thank you.

PR #727, opened 2020-11-26
PHP 8.x support

### LINKED ISSUE #725: Provide PHP 8.x support
Currently, the SDK provides support for PHP version 7.x but it doesn't for the upcoming 8.x due to a composer restriction:

https://github.com/mailgun/mailgun-php/blob/a084abdb8601c0380b668651c169e0560fb435d0/composer.json#L5

An update of the `composer.json` file to allow PHP 8.x also requires an update of the PHP Unit version, as the current one is set for 7.5, that is also restricted to PHP version 7.x.

https://github.com/mailgun/mailgun-php/blob/a084abdb8601c0380b668651c169e0560fb435d0/composer.json#L13

https://packagist.org/packages/phpunit/phpunit#7.5.20

Upgrading the SDK for this newer PHP version might imply a BC Break for previous PHP versions, so maybe is worth creating a new major release as we can drop support for PHP 7.1 and 7.2.

DavidGarciaCat:
Closes #725

----

Before merging this new PR, we will wait for https://github.com/mailgun/mailgun-php/pull/726 to be merged so we can provide a new SDK version with the highly requested Email Validator feature.

This PR to enable PHP 8.0 support causes a BC Break so that it will be flagged as SDK 4.0.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,17 +3,17 @@
     "description": "The Mailgun SDK provides methods for all API functions.",
     "require": {
         "php": "^7.3 || ^8.0",
-        "psr/http-client": "^1.0,>=1.0.1",
-        "php-http/multipart-stream-builder": "^1.0,>=1.1.2",
-        "php-http/client-common": "^2.2,>=2.2.1",
-        "php-http/discovery": "^1.9,>=1.9.1",
-        "webmozart/assert": "^1.9,>=1.9.1"
+        "psr/http-client": "^1.0.1",
+        "php-http/multipart-stream-builder": "^1.1.2",
+        "php-http/client-common": "^2.2.1",
+        "php-http/discovery": "^1.9.1",
+        "webmozart/assert": "^1.9.1"
     },
     "require-dev": {
         "phpunit/phpunit": "^9.3",
-        "php-http/guzzle7-adapter": "^0.1,>=0.1.1",
-        "nyholm/psr7": "^1.3,>=1.3.1",
-        "nyholm/nsa": "^1.2,>=1.2.1"
+        "php-http/guzzle7-adapter": "^0.1.1",
+        "nyholm/psr7": "^1.3.1",
+        "nyholm/nsa": "^1.2.1"
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -2,18 +2,18 @@
     "name": "mailgun/mailgun-php",
     "description": "The Mailgun SDK provides methods for all API functions.",
     "require": {
-        "php": "^7.1",
-        "psr/http-client": "^1.0",
-        "php-http/multipart-stream-builder": "^1.0",
-        "php-http/client-common": "^1.9 || ^2.0",
-        "php-http/discovery": "^1.6",
-        "webmozart/assert": "^1.6"
+        "php": "^7.3 || ^8.0",
+        "psr/http-client": "^1.0,>=1.0.1",
+        "php-http/multipart-stream-builder": "^1.0,>=1.1.2",
+        "php-http/client-common": "^2.2,>=2.2.1",
+        "php-http/discovery": "^1.9,>=1.9.1",
+        "webmozart/assert": "^1.9,>=1.9.1"
     },
     "require-dev": {
-        "phpunit/phpunit": "^7.5",
-        "php-http/guzzle6-adapter": "^1.0 || ^2.0",
-        "nyholm/psr7": "^1.0",
-        "nyholm/nsa": "^1.1"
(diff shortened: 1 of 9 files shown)


composer.json
```diff
@@ -2,18 +2,18 @@
...
-        "php": "^7.1",
-        "psr/http-client": "^1.0",
-        "php-http/multipart-stream-builder": "^1.0",
-        "php-http/client-common": "^1.9 || ^2.0",
-        "php-http/discovery": "^1.6",
-        "webmozart/assert": "^1.6"
+        "php": "^7.3 || ^8.0",
+        "psr/http-client": "^1.0,>=1.0.1",
+        "php-http/multipart-stream-builder": "^1.0,>=1.1.2",
+        "php-http/client-common": "^2.2,>=2.2.1",
```
Nyholm, line 8: This would be the same
```suggestion
        "php-http/client-common": "^2.2.1",
```

Nyholm: But why do you need to specify the patch version?

Nyholm:
We could keep supporting PHP 7.1 and use symfony/phpunit-bridge to ease the testing

composer.json
DavidGarciaCat: Because the patch versions are the ones adding support for PHP 8, the previous one still has a direct link to `^7.x` instead of `>=7.x`

Nyholm: But let composer handle what exact versions to use. As long as "some" version of the dependency support PHP8 then it is fine.

DavidGarciaCat: OK, I will re-adjust it as part of the next commits, as this is still a draft. Thanks for the feedback!

lptn:
any news guys?

repat:
I'd also be interested

DavidGarciaCat:
Still WIP, but I will try to invest some time this month.

lptn:
Hey @DavidGarciaCat

How about a simple migration without BC? It's always good for the whole ecosystem to support latest PHP version. I don't see any BC changes in a diff, only PHP version changes

DavidGarciaCat:
Hi @lptn 

According to the Packagist website (repository for `composer`):
- `phpunit#7.5.20` => `php: ^7.1` https://packagist.org/packages/phpunit/phpunit#7.5.20
- `phpunit#8.0.0` => `php: ^7.2` https://packagist.org/packages/phpunit/phpunit#8.0.0
- `phpunit#8.5.14` => `php: >=7.2` https://packagist.org/packages/phpunit/phpunit#8.5.14
- `phpunit#9.0.0` => `php: ^7.3` https://packagist.org/packages/phpunit/phpunit#9.0.0
- `phpunit#9.5.1` => `php: >=7.3` https://packagist.org/packages/phpunit/phpunit#9.5.1

The problem is the SDK supports `PHP ^7.1` as it's `composer` requirement, that means we need to keep using PHPUnit 7.x as the newer versions depend on newer PHP versions too.

And the main blocker is that from PHPunit 7 yo 8, the `setUp()` and `turnDown()` methods (that are the most common, but other PHPunit methods too) are now type-hinted as a mandatory requirement (when it was not the case before).

So:
- given PHPUnit 7.5 doesn't expect support for `>= 7.1` but `^ 7.1` (that limits to 7.y.z), then we cannot use it for PHP 8
- given PHPUnit 8.0 expects a minimum version of PHP 7.2, then we can't keep the compatibility with PHP 7.1
- the same problem happens with the newer versions, as they expect newer PHP versions too

So, unless my brain doesn't work anymore and I can't remember a way to write the PHPunit tests keeping compatibility between those versions, I can't see a way to upgrade the PHP version without updating the tests

stephanvierkant:
Let's drop support for 7.1 since it's End of Life for more than a year. Even it's successor 7.2 is EoL already.

Nyholm:
I'll have a look at this PR this week. David and I will make sure to get PHP8 support this month.

composer.json
DavidGarciaCat: I've implemented this change locally. I'll push it soon to the repo.

Nyholm:
My general rule is: Keep supporting old PHP versions unless there is a problem. 

I see that the current state of the PR is just bumping PHP versions and some dependencies. Our BC promise does not include our test code. 

This PR is safe to merge. Thank you @DavidGarciaCat

stephanvierkant:
Thanks! Will this be released as 4.0.0 or 3.3.0? And are there other issues/PR's blocking a release?

Nyholm:
It will be 3.3. I'll wait for David to do some reviews. I'll prepare a release shortly

DavidGarciaCat:
Tagged version `3.3.0` is now available with PHP 8 support

PR #739, opened 2021-01-12
Add additional events to TotalResponseItem

Arkitecht:
Currently the TotalResponseItem class does not parse opened, clicked, unsubscribed, and stored events.
This PR solves that.

It also adds tests for the new events, and adds testing of totals in the response

### DIFF
--- a/tests/Api/StatsTest.php
+++ b/tests/Api/StatsTest.php
@@ -42,6 +42,22 @@ class StatsTest extends TestCase
         $this->assertInstanceOf(TotalResponse::class, $total);
         $this->assertCount(count($responseData['stats']), $total->getStats());
         $this->assertContainsOnlyInstancesOf(TotalResponseItem::class, $total->getStats());
+
+        $event = $queryParameters['event'];
+        $responseStat = $total->getStats()[0];
+        $statGetter = 'get'.ucwords($event);
+
+        if ('failed' !== $event) {
+            $expectedTotal = $responseData['stats'][0][$event]['total'];
+            $actualTotal = $responseStat->$statGetter()['total'];
+        }
+
+        if ('failed' === $event) {
+            $expectedTotal = $responseData['stats'][0][$event]['permanent']['total'];
+            $actualTotal = $responseStat->$statGetter()['permanent']['total'];
+        }
+
+        $this->assertEquals($expectedTotal, $actualTotal);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tests/Api/StatsTest.php
```diff
@@ -42,6 +42,22 @@ public function testTotal($queryParameters, $responseData)
...
+        $event = $queryParameters['event'];
+        $responseStat = $total->getStats()[0];
+        $statGetter = 'get'.ucwords($event);
+
+        if ('failed' !== $event) {
+            $expectedTotal = $responseData['stats'][0][$event]['total'];
+            $actualTotal = $responseStat->$statGetter()['total'];
+        }
+
+        if ('failed' == $event) {
```
DavidGarciaCat, line 55: A trivial point, buy maybe `===` is better here, as we intend to get a match

```suggestion
        if ('failed' === $event) {
```

tests/Api/StatsTest.php
```diff
@@ -109,6 +125,32 @@ public function totalProvider()
                     ],
                 ]),
             ],
+            'clicked events' => [
```
DavidGarciaCat, line 128: I can see you have added to this payload both the `clicked` and `opened` events, but I cannot see either the `unsubscribed` or `stored`. Are they already here and I didn't notice them?

DavidGarciaCat (commented):
Hello @Arkitecht 

Thank you for your contribution.

I have added a couple of minor comments. If you can spare a few minutes to fix them, then I am more than happy to merge it and tag these new changes as part of the next release.

Thank you,

D

tests/Api/StatsTest.php
Arkitecht: Sure, not a problem, I will push that change up shortly. 
Thanks for reviewing!

tests/Api/StatsTest.php
Arkitecht: I initially just added the clicked and opened, but added the others now.

Arkitecht:
@DavidGarciaCat - I made the requested changes, please let me know if you see anything else that needs updating.

DavidGarciaCat (approved):
Everything seems good to me. Thanks for your contribution.

DavidGarciaCat:
It looks like the required builds are not being executed, so please let me check it with another contributor who has more permissions so he can update the settings for GitHub actions and we can merge it.

Nyholm (approved):
Awesome. Thank you

PR #778, opened 2021-08-18
Update dev dependencies

Nyholm:
The `php-http/guzzle7-adapter` is not needed anymore. We can use guzzle or symfony's http client instead. 

Disclaimer: I maintain all three packages.


DavidGarciaCat (approved):
I knew you were maintaining some of those vendors, but not all of them!

PR #780, opened 2021-08-18
Update links and remove Buzz example

Nyholm:
Buzz is a great client, but I rather recommend using Symfony. 

One can argue that we should recommend Guzzle instead but Symfony is way more well maintained. 

Disclaimer: I am part of the team maintaining Symfony, the team for Guzzle and Im also maintaining Buzz.


DavidGarciaCat (approved):

Nyholm:
Thank you for merging

Issue #783, opened 2021-08-20
Webhook is still using ClientInterface type declaration


ixarlie:
The Webhook class which is extending from HttpApi is still using the Psr\Http\Client\ClientInterface type declaration.

This is raising an exception, so as it was done for its parent class HttpApi, the type declaration should be less strict.
```
"Argument 1 passed to Mailgun\Api\Webhook::__construct() must implement interface
Psr\Http\Client\ClientInterface, instance of Http\Client\Common\PluginClient given,
called in .../vendor/mailgun/mailgun-php/src/Mailgun.php on line 139"
```

Relates https://github.com/mailgun/mailgun-php/issues/642

Mailgun\Api\Suppression is using the same type declaration Psr\Http\Client\ClientInterface, but I am not sure if this class should be updated as well.

Thanks

oleksandr-mykhailenko:
Fixes were made

Issue #577, opened 2019-02-28
Can not get stored email attachment using default hydrator.


kormik:
It is not possible to get stored email attachment using default ModelHydrator.
I used this:
`$mailgunClient->attachment()->show($url);`
and I got this Exception:
`The ModelHydrator cannot hydrate response with Content-Type: image/jpeg`

Even though there is a model class for attachment it is not possible to hydrate it because the `ModelHydrator` always expect JSON but endpoint for attachments returns actual file content.

DavidGarciaCat:
Hi @kormik 

Can you please provide an example for this case?

Do we just need to send an email with an attached file and then trying to retrieve it, and that's all?

Thanks,

kormik:
I think so. You can test it on stored email with an attachment.
- When you retrieve the email, you will get an url for the attachment.
- When you try to retrieve the attachment using provided url you will get the exception.
Endpoint for attachments looks like:
`https://storage.eu.mailgun.net/v3/domains/<OUR_DOMAIN>/messages/<MESSAGEID>/attachments/0`
Which returns attachment content but ModelHydrator always expects json.

DavidGarciaCat:
Thanks @kormik 

We'll try to take a look and fix it.

If, in the meantime, you want to expedite a potential solution, PRs are more than welcome.

Thanks,

PR #784, opened 2021-08-20
Remove type declaration for Webhooks and Supression classes

ixarlie:
Resolves https://github.com/mailgun/mailgun-php/issues/783


src/Api/Suppression.php
```diff
@@ -11,6 +11,7 @@
 
 namespace Mailgun\Api;
 
+use Http\Client\Common\PluginClient;
```
Nyholm, line 14: Lets use `Http\Client\HttpClient` instead.

DavidGarciaCat (commented):
There is also a BC break - we should confirm it won't cause any problem (because we require it to return a successful status)

ixarlie:
Hi @DavidGarciaCat , yeah I noticed that there are BC breaks too but I am not sure how to resolve it.

My project was using the version 2.x and we were having this same issue https://github.com/mailgun/mailgun-php/issues/577 . The version 2.x is using Http\Client\HttpClient instead of Psr\Http\Client\ClientInterface so everything was working well in regards to the http client.

This fix https://github.com/mailgun/mailgun-php/pull/590 is not resolved in 2.8.1, so we decided to upgrade to "3.2.x" (we do have some other dependencies conflicts right now so we cannot go further)

The version 3.2.x requires `"php-http/client-common": "^1.9 || ^2.0" `(my project is locked at `php-http/httplug v1.1.0`) but in the php-http/httplug 1.x the interface `Http\Client|HttpClient` is not extending `Psr\Http\Client\ClientInterface` as it does in 2.x.

I'd say this issue is not happening for mailgun-php versions > 3.2.x because from 3.3.0 composer is `php-http/client-common: ^2.2.1` which requires `php-http/httplug: ^2.0`

However, it is odd that the HttpApi is doing this logic and there is a BC in Mailgun\Api\Webhook when Webhook is extending HttpApi. Suppression is different because is not extending from HttpApi, so that's why I copied the same bits from HttpApi.
```php
    // https://github.com/mailgun/mailgun-php/blob/master/src/Api/HttpApi.php
    public function __construct($httpClient, RequestBuilder $requestBuilder, Hydrator $hydrator)
    {
        if (!is_a($httpClient, ClientInterface::class) &&
            !is_a($httpClient, PluginClient::class)) {
            throw new \RuntimeException('httpClient must be an instance of
            Psr\Http\Client\ClientInterface or Http\Client\Common\PluginClient');
        }
```
BTW, I think the code above is not necessary for versions greater than 3.3.0 and the type declaration can be restored to ClientInterface because from 3.3.0 as I commented before httplug is 2.x and PluginClient implements Http\Client|HttpClient which extends Psr\Http\Client\ClientInterface.


Probably we may not really need this PR and the linked issue can be closed, but then I really need https://github.com/mailgun/mailgun-php/pull/590 to be pushed in the 2.x branch and a new release 2.8.2 or 2.9.0 please.


Thanks in advanced!

DavidGarciaCat:
Hey @Nyholm and @Pavlico 

What do you guys think might be better here? Moving that #590 to 2.x (as @ixarlie suggested) or keep moving forward with this PR?

Personally, I would like to reduce the number of versions we might need to maintain, but I would like to hear your point of view too.

Cheers,

DavidGarciaCat:
And @ixarlie 

After checking the failing analysis here (https://github.com/mailgun/mailgun-php/pull/784/checks?check_run_id=3388496952) it seems to be quite clear the problem:

```
[BC] CHANGED: The parameter $httpClient of Mailgun\Api\Webhook#__construct() changed from Psr\Http\Client\ClientInterface to no type
[BC] CHANGED: The parameter $httpClient of Mailgun\Api\Suppression#__construct() changed from Psr\Http\Client\ClientInterface to no type
```

I believe (and I am not sure) it might be fixed if we run a conditional (such as `$client instanceof ...`) but we should test it.

Cheers,

Nyholm:
The BC checking script is very strict. 
There is actually no scenario where a users code will break if we merge this. 

I’ll be happy to merge this pr and then tag a minor release.

DavidGarciaCat:
> The BC checking script is very strict.
> There is actually no scenario where a users code will break if we merge this.
> 
> I’ll be happy to merge this pr and then tag a minor release.

Works for me. But my merge permissions don't allow me to merge skipping that check... Maybe you can help here?

src/Api/Suppression.php
```diff
@@ -41,8 +42,13 @@ class Suppression
      */
     private $hydrator;
 
-    public function __construct(ClientInterface $httpClient, RequestBuilder $requestBuilder, Hydrator $hydrator)
+    public function __construct($httpClient, RequestBuilder $requestBuilder, Hydrator $hydrator)
     {
+        if (!is_a($httpClient, ClientInterface::class) &&
```
Nyholm, line 47: Let's change this to `instanceof` instead.

src/Api/Suppression.php
```diff
@@ -41,8 +42,13 @@ class Suppression
      */
     private $hydrator;
 
-    public function __construct(ClientInterface $httpClient, RequestBuilder $requestBuilder, Hydrator $hydrator)
+    public function __construct($httpClient, RequestBuilder $requestBuilder, Hydrator $hydrator)
```
Nyholm, line 45: Lets add a doc block that describes `$httpClient`. Same on `Webhook`

Nyholm (changes requested):
Thank you. Only two small changes.

src/Api/Suppression.php
ixarlie: Hi @Nyholm , should I do the same for `HttpApi` ?

https://github.com/mailgun/mailgun-php/blob/master/src/Api/HttpApi.php#L47-L53

Nyholm: Yes please

src/Api/Suppression.php
ixarlie: The doc block was added

ixarlie:
More failing checks than before :confused:

Nyholm:
Thank you. 
I’m fine with the pr in its current state. I’ll merge it next time I’m at a computer. I’ll deal with updating the baseline and the failing checks.

Pavlico:
Hi @Nyholm, should I try to add fixes to code so it passes the thingy or as you said you will merge it in this state :)

DavidGarciaCat:
> Hi @Nyholm, should I try to add fixes to code so it passes the thingy or as you said you will merge it in this state :)

Hey @Pavlico 

If you have the time to work on it, then yes please!

Thank you!

Pavlico:
Moved to #798

PR #786, opened 2021-08-23
Add template support when using BatchMessage

kochjoel:
Add function to set the template and only throw the error message when text, html and template variables are missing.
Also mention "setTextBody", "setHtmlBody" and the new "setTemplate" in the readme.

Before this, you had to use "setMessage" function to set all the variables including from, to, template, text/html etc if you wanted to use templates.


kochjoel:
@DavidGarciaCat The issues are now fixed and all of the tests have passed.

gelform:
+1 to see this PR get committed.

Pavlico:
All good for me, @DavidGarciaCat agreed?

Issue #790, opened 2021-09-13
Default value of description of mailing list


quallrum:
Default value of description argument of `Api\MailingList::create()` is currently `null`, which is incompatible with `MultipartStreamBuilder::create()` method from package `php-http/multipart-stream-builder` that doesn't accept null values. That leads to an exception when I try to create a mailing list without description, but, according to the Mailgun's documentation, `description` field is optional. Probably a better way is to avoid adding null arguments to request body in `Api\MailingList::create()` or removing them from array using `array_filter()` function.

php-http/multipart-stream-builder 1.2.0
mailgun/mailgun-php 3.5.1

PR #791, opened 2021-09-15
Fix null params

Pavlico:


Pavlico:
#790

DavidGarciaCat (commented):
Shouldn't we add a PHPUnit test here to confirm the different behaviours when providing (or not providing) these nullable parameters?

Pavlico:
@DavidGarciaCat added

DavidGarciaCat (approved):
Thank you!

PR #797, opened 2021-10-06
Added tests and method for adding content with string

Pavlico:


Pavlico:
@DavidGarciaCat can you reviw :)

PR #798, opened 2021-10-08
Issue 783 phpstan fixes

Pavlico:


src/Api/HttpApi.php
```diff
@@ -44,13 +43,13 @@ abstract class HttpApi
      */
     protected $requestBuilder;
 
-    public function __construct($httpClient, RequestBuilder $requestBuilder, Hydrator $hydrator)
+    /**
+     * @param ClientInterface $httpClient     Client used to send the requests
+     * @param RequestBuilder  $requestBuilder Builder that creates request based on params
+     * @param Hydrator        $hydrator       Transforms request body into array
```
ZebulanStanphill, line 49: Since the types are already defined in the code, it seems redundant to repeat them in the docs here if we aren't going to make them more specific. I think the following would work:
```suggestion
     * @param $httpClient     Client used to send the requests
     * @param $requestBuilder Builder that creates request based on params
     * @param $hydrator       Transforms request body into array
```
(And ditto for the other doc-blocks in this PR.)

src/Api/HttpApi.php
```diff
@@ -30,7 +29,7 @@ abstract class HttpApi
     /**
      * The HTTP client.
      *
-     * @var ClientInterface|PluginClient
+     * @var ClientInterface
```
ZebulanStanphill, line 32: In case anyone reviewing this PR is confused, `PluginClient` implements `HttpClient` which implements `ClientInterface`, so the type union here was redundant, and it makes sense to simplify the type here to just `ClientInterface`.

src/Api/HttpApi.php
DavidGarciaCat: The `@param TypeHint $variableName` is redundant, but it also breaks the PSR, as far as I can remember. When adding the `@param` we should also add a description to explain what that variable is.

Consider something like `@param GenericRepositoryInterface $repo` - It's redundant, but you might want to describe what you expect from that repo to be passed.

In my opinion, keeping it is not problematic at all.

ZebulanStanphill: I looked it up, and you're right: omitting the type from an `@param` does break standards, namely the [PSR-19 draft](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#510-param). Thanks for pointing that out.

PR #801, opened 2021-10-13
Fix some missing return/throw types.

ZebulanStanphill:
I've been trying to fix all the type errors reported by PHPStan and Psalm, but it turns out there's a LOT, so to make things easier to review, I'm splitting the changes up into several PRs, starting with this one.

This PR adds some missing return/throw types, clarifies some existing ones, and moves some from PHPDocs to native type declarations. I've mostly just done the simple ones here. I'll do the more complex ones (basically everything involving `$this->hydrateResponse`) in a later PR.

### DIFF
--- a/src/Hydrator/NoopHydrator.php
+++ b/src/Hydrator/NoopHydrator.php
@@ -24,6 +24,7 @@ final class NoopHydrator implements Hydrator
      * @param class-string $class
      *
      * @throws \LogicException
+     * @return never
      */
     public function hydrate(ResponseInterface $response, string $class)
     {
--- a/src/Hydrator/NoopHydrator.php
+++ b/src/Hydrator/NoopHydrator.php
@@ -24,6 +24,7 @@ final class NoopHydrator implements Hydrator
      * @param class-string $class
      *
      * @throws \LogicException
+     *
      * @return never
      */
     public function hydrate(ResponseInterface $response, string $class)
--- a/src/Api/Attachment.php
+++ b/src/Api/Attachment.php
@@ -12,11 +12,9 @@ declare(strict_types=1);
 namespace Mailgun\Api;
 
 use Mailgun\Assert;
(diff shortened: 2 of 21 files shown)


src/Hydrator/NoopHydrator.php
```diff
@@ -24,6 +24,8 @@ final class NoopHydrator implements Hydrator
      * @param class-string $class
      *
      * @throws \LogicException
+     *
+     * @return never
```
DavidGarciaCat, line 28: It seems unnecessary to me

DavidGarciaCat (approved):
LGTM - Thanks!

src/Hydrator/NoopHydrator.php
ZebulanStanphill: This helps Psalm and PHPStan recognize that not only does this function throw, but it specifically _always_ throws. So if you put any code after a call to this function, the static analyzers will know there's something wrong, because the code would never be executed. PHP 8.1 is actually adding a [native `never` return type](https://php.watch/rfcs/noreturn_type) for this reason.

DavidGarciaCat: I understand, but this SDK is based on PHP 7.3 or greater, which means we should not use code that has been specifically added (or will be added) in future versions.

ZebulanStanphill: Well, since it's just a comment annotation (rather than a native return type) it has no effect on the code's PHP 7.3 compatibility. But Psalm and PHPStan already support this annotation, so it will provide static analysis to everyone using those tools regardless of which PHP version they're using.

DavidGarciaCat (commented):
We have a BC failure here - can we please review this case so we can move forward with the update?

https://github.com/mailgun/mailgun-php/pull/801/checks?check_run_id=3883732923

PR #805, opened 2021-10-25
Update Job.php

simon-at-fugu:
Fix "Undefined index: download_url" PHP ERROR -> Wrap with isset().


oleksandr-mykhailenko:
This was done already, Thank you

PR #815, opened 2021-12-21
Message\ShowResponse set empty array when no message headers key

ixarlie:
This PR resolves an issue when a `Message\ShowResponse` is created and the argument `$data` does not have a `message-headers` key.

`public function getMessageHeaders(): array`

In that case, making use of `getMessageHeaders` a TypeError is thrown.

> Return value of Mailgun\Model\Message\ShowResponse::getMessageHeaders() must be of the type array, null returned

Using the same approach from the "attachments" property.


DavidGarciaCat (approved):
Nice catch!

Pavlico:
@ixarlie hi can you fix composer issue, its simple change in composer.json file

ixarlie:
> @ixarlie hi can you fix composer issue, its simple change in composer.json file

I'm sorry @Pavlico , not sure what kind of fix you mean. Looking at the composer.json file but no idea what to do.
If you can give me some directions. Thanks!

Pavlico:
We cool, I think you have old composer on your branch that doesn't pass this check, update your local master with current version and then merge it into your branch and it will fix it, I think :) or if you don't want to update master just copy current composer.json file to your branch, you can see details on how your composer.json file is different in composer normalise check

ixarlie:
@Pavlico that worked! Thank you!

Pavlico:
Thanks to you too ! :)

PR #818, opened 2022-02-02
Bugfix: Email Validations v3 / Add Email Validations to documentation.

mikeundefined:
We received some tickets asking how to perform a email validation with the mailgun-php library due to lack of documentation. Added some examples to the documentation for user convenience :)

While testing the classes out, I discovered the v3 class was broken due to an API url change. (Must include v3 in the url now for email validation v3)

I made the correction. Thanks for reviewing! <3

- Always open to feedback.


DavidGarciaCat (commented):
Thanks for your PR

Question: given all the documentation sets the API v4, why do you think we should downgrade?

https://documentation.mailgun.com/en/latest/api-email-validation.html

TheNando (approved):

michaelhyattdev:
There is a new class called EmailValidationV4, however there might still be quite a few people using v3. There are two different types of API keys on a Mailgun account. The private API key and the public validation key.

The public validation key allows for the user to set a limit on validations. The private api key does not. As far as I am aware, the public validation key only works on v3 while v4 requires the use of the private api key.

I use v3 on my own site just in case someone gets happy with one of my sign up forms :) They will get cut off by my public validation limit and save me $$.

So I think we should keep both classes for now :).

DavidGarciaCat:
Understood, but in that case, I have the feeling the PR is incomplete because the Email Validation has 2 potential API endpoints but just one of them was updated. You can see it by viewing the whole file:

https://github.com/mailgun/mailgun-php/blob/9a8a9773b9f9f8cb02148cd105c2560f6bb4ba74/src/Api/EmailValidation.php

Would you mind please updating the 2nd endpoint and its test, please?

DavidGarciaCat:
It might also be a good idea to [trigger an error](https://www.php.net/trigger_error) and set the deprecation level when using the V3 endpoints. Aiming not to cause a blocker, you could "silence" the error adding the `@` prefix (in the same way that many frameworks do it).

michaelhyattdev:
Great catch sir. I really appreciate the time you've invested to look over this. I have updated the parse endpoint and it's test.

I know for the time being several endpoints using v3 are not deprecated yet such as the endpoints for the messages, ips, and events apis.

oleksandr-mykhailenko (approved):

PR #819, opened 2022-02-16
Use webhook signing key instead of API key for webhook signature verification

damiantw:
When instantiating a `Mailgun` instance you are expected to provide an API key that is used to authenticate requests to the REST API.

```php
// First, instantiate the SDK with your API credentials
$mg = Mailgun::create('key-example');
```

This API key was passed to the `Api\Webhook` instance that is instantiated via `$mg->webhooks()`.

```php
public function webhooks(): Api\Webhook
{
    return new Api\Webhook($this->httpClient, $this->requestBuilder, $this->hydrator, $this->apiKey);
}
```

https://github.com/mailgun/mailgun-php/blob/master/src/Mailgun.php#L142

This passed API key is used in the [`verifyWebhookSignature`](https://github.com/mailgun/mailgun-php/blob/master/src/Api/Webhook.php#L55) function. 

This will result in improper verification results as the account webhook signing key should be used to compute the signature, not the REST API key.

As is, you must manually instantiate an `Api\Webhook` instance to provide a webhook signing key and properly validate webhook signatures using this library.

```php
$webhook = new Webhook((new HttpClientConfigurator)->createConfiguredClient(), new RequestBuilder, new ModelHydrator,  'signing-key');
        
$webhook->verifyWebhookSignature($timestamp, $token, $signature); // This will work
```

This PR allows for passing a webhook signing key using the `$mg->webhooks('signing-key')` function.

### DIFF
--- a/src/Api/Webhook.php
+++ b/src/Api/Webhook.php
@@ -32,12 +32,12 @@ class Webhook extends HttpApi
     /**
      * @var string
      */
-    private $apiKey;
+    private $signingKey;
 
-    public function __construct(ClientInterface $httpClient, RequestBuilder $requestBuilder, Hydrator $hydrator, string $apiKey)
+    public function __construct(ClientInterface $httpClient, RequestBuilder $requestBuilder, Hydrator $hydrator, string $signingKey)
     {
         parent::__construct($httpClient, $requestBuilder, $hydrator);
-        $this->apiKey = $apiKey;
+        $this->signingKey = $signingKey;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


faizanakram99 (changes requested):

dpslwk:
changing  to 
`return new Api\Webhook($this->httpClient, $this->requestBuilder, $this->hydrator, $signingKey ?: $this->apiKey);`
would allow `$signingKey` to be optional and keep the old behaviour
which would then not be a behaviour BC, but does still leave a question of if the signature change is classed as a BC

bizurkur:
I "love" how this was recognized as a BC break, yet still released as a bug fix.

Please read how [proper versioning](https://semver.org/) should take place. If you break the API, it's not a bug fix.

Issue #825, opened 2022-06-10
BatchMessage generates "invalid" JSON for recipients with an empty set of recipient-variables


dbaeck:
According to mailgun api docs, recipient variables should be sent as a json dictionary with entries of the form `"mail-address":{...}` (https://documentation.mailgun.com/en/latest/user_manual.html#batch-sending-1).

Since a few days, mailgun seems to apply different validation rules on this json dictionary, and no longer allows `"mail-address":[]` for empty entries, which is also valid json and worked until recently. `"mail-adress": {}` is totally fine.

The problem now is, mailgun-php would encody an empty `$variables` array as `[]`. Also, `$batchMessage->setFromAddress(...)` adds an entry to the `recipient-variables`. So setting the sender address without variables (which would normaly be the case), leads directly to invalid `recipient-variables` and an API error (400) on sending.

Maybe `JSON_FORCE_OBJECT` when encoding would fix this problem already.

deviarte:
We are getting the same issue. Impacting many sends!

oleksandr-mykhailenko:
Changes were merged in another PR. Thank you

PR #826, opened 2022-06-10
BatchMessage generates "invalid" JSON

deviarte:
The Mailgun messages API endpoint now seems to expect that all JSON provided keys have values or empty object.

Currently, empty keys are converted to arrays. This change forces it to be a JSON object.

Reference: https://github.com/mailgun/mailgun-php/issues/825


oleksandr-mykhailenko:
@deviarte Your proposed changed were implemented here - https://github.com/mailgun/mailgun-php/pull/829
Thank you for contributing

Issue #827, opened 2022-06-13
Undefined index: assignable_to_pools - Ip\IndexResponse


TWithers:
When using the following code:

`$mg->ips()->domainIndex('my.domain.mg');`

I get the warning:

`PHP Notice:  Undefined index: assignable_to_pools`

It looks like the IPs API is not returning that field in all instances, and probably should be changed to a null coalescing operator inside Mailgun\Model\Ip\IndexResponse on line 45:

`$model->assignableToPools = $data['assignable_to_pools'] ?? [];`

oleksandr-mykhailenko:
Proposed changes were merge in another PR. Thank you

PR #828, opened 2022-06-13
Use null coalescing operator in IndexResponse.php

TWithers:
In reference to Issue #827


oleksandr-mykhailenko:
Hello @TWithers 
Your proposed changes were merged. Thank you =

PR #864, opened 2023-06-18
close resources even if request failed

kima92:
handle cases when Mailgun is used inside a long-running process (queue worker for example), catch the exception without closing the resources


oleksandr-mykhailenko:
Thank you
In the latest release we implemented your changes

PR #880, opened 2023-11-14
composer.json - upgrade min php to 8.1

matteotrubini:


oleksandr-mykhailenko:
Hello @matteotrubini 
We support both version, because some libraries have non compatible changes

matteotrubini:
So README should be corrected: "minimum php version 8.1"

oleksandr-mykhailenko:
> So README should be corrected: "minimum php version 8.1"

Yes, you are right. We will fix it

PR #883, opened 2023-11-29
Better json_decode exception handling

### LINKED ISSUE #882: Bug: Proper json_last_error() handling with try catch
So, the last release introduces [more strict JSON syntax check](https://github.com/mailgun/mailgun-php/blob/888d70f20c93278ea525c448d88a9f114b5649ae/src/Hydrator/ModelHydrator.php#L39C40-L39C40).

There is an inherent/global issue with how json_last_error() works. In short, the returned error code can be a false positive. It does not reset after a successful encode or decode call. It changes only when there is a new error code.

At the issue core is the following sequence, which gives unintuitive results: 
```php
<?php
json_decode("\00invalid json");
var_dump(json_last_error());

json_decode("[]", true, 512, JSON_THROW_ON_ERROR);
var_dump(json_last_error());
```
**Both** var dumps would return an error code. You can read about the issue source over this thread: https://github.com/php/php-src/issues/10166

While in isolation this works ok, in the integration it becomes unreliable if, in the same call/request, there was another error code thrown (i.e. to check if a string is a json). I'm preparing a PR that uses try-catch to handle the error properly.

Pilskalns:
Wrap json_decode in proper handling for strict syntax checks.

Fixes #882

### DIFF
--- a/src/Hydrator/ArrayHydrator.php
+++ b/src/Hydrator/ArrayHydrator.php
@@ -33,9 +33,10 @@ final class ArrayHydrator implements Hydrator
             throw new HydrationException('The ArrayHydrator cannot hydrate response with Content-Type:'.$response->getHeaderLine('Content-Type'));
         }
 
-        $content = json_decode($body, true, 512, JSON_THROW_ON_ERROR);
-        if (JSON_ERROR_NONE !== json_last_error()) {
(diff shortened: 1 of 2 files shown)


wintrue:
There are still mistakes.

PR #894, opened 2024-02-05
fix documentation URLs

Gavrisimo:
I found that URLs with out `/en/latest` are resulting in 404 errors:

- https://documentation.mailgun.com/api_reference.html
- https://documentation.mailgun.com/api-domains.html
- https://documentation.mailgun.com/api-routes.html
- https://documentation.mailgun.com/api-suppressions.html

Just adding this missing `/en/latest` fixes them.

I also fixed the https://documentation.mailgun.com/user_manual.html#um-spam-filter although that link seems to be incorrectly placed there, I'm not really sure, you'll guys have to take a look at that one.

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -1,7 +1,7 @@
 # API documentation
 
 This page will document the API classes and ways to properly use the API. These resources will eventually move to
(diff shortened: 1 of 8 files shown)


oleksandr-mykhailenko (approved):
You're right, Thank you

PR #914, opened 2024-07-30
De 1301 mailgun sdk the template search filters do not match the documentation

oleksandr-mykhailenko:


src/Api/MailingList/Member.php
```diff
@@ -156,11 +156,11 @@ public function createMultiple(string $list, array $members, bool $upsert = fals
                         if (is_array($value)) {
                             $value = json_encode($value);
                         }
-                        // We should assert that "vars"'s $value is a string.
+                        break;
+                    // We should assert that "vars"'s $value is a string.
                         // no break
                     case 'name':
```
CadenP, line 162: Just checking: should this `break` have been added? There's a comment two lines below that says "no break", and the comment above that seems to explain why. `json_encode` could return `false`, so perhaps that's what was being avoided here.

I haven't read through any of this code, so I don't know what it does, and I'm not using this version, so I don't know whether it breaks or improves anything. I just happened across this and thought I'd point it out and ask.

PR #928, opened 2025-02-27
Add useAutomaticSenderSecurity support in DomainV4

matteotrubini:
As per title.


matteotrubini:
Thanks for the merge. I just noticed that the indentation inside `updateUseAutomaticSenderSecurity()` is missing. I’m sorry for the oversight.

PR #934, opened 2025-07-02
Use subaccount model in create API response

svenluijten:
This PR ensures the created subaccount is available to consumers of this API without having to make a new request. According to [the docs](https://documentation.mailgun.com/docs/mailgun/api-reference/openapi-final/tag/Subaccounts/#tag/Subaccounts/operation/post-v5-accounts-subaccounts), the `POST /v5/accounts/subaccounts` returns the created subaccount, so we'll use the already existing model in the project to return it.


svenluijten:
It should be noted that this is a breaking change, as the `$message` property was previously filled with the response. But it made no sense that I had to call `->getMessage()` to get the _response body_. I figured a separate `->getItem()` method that returns the domain model works better.

Issue #948, opened 2026-03-09
Please update webmozart/assert requirement


aaronbauman:
Drupal 11 requires PHP 8.4

I thought updating webmozart/assert would be sufficient, but it's just the tip of the iceberg.

Current webmozart/assert dependency is 4 years out of date, and preventing me from running alongside latest release of https://github.com/simplesamlphp/saml2-legacy

Looks like none of the tooling has been updated in that 4 years, so the entire test chain needs to be updated as well.

And 8.4 will require breaking changes, so either 8.4 needs fully separate tests, or support for earlier PHP versions needs to be dropped.

CircuitCipher:
This is a blocker for anyone trying to use the Mailgun module on any recent versions of Drupal (https://www.drupal.org/project/mailgun/issues/3580037)

aaronbauman:
My struggles to get this working are on the 2 linked PRs, and I don't think it's worth the effort to include support for two incompatible versions of the webmozart/assert package.

Seems like a major version release for mailgun-php would be appropriate, dropping support for PHP7.

This would allow significant reduction in the tooling and testing coverage, as well as updates for packages like webmozart/assert.

I'd be happy to work on such a PR, but will await feedback from maintainers before I sink any more time on this.

oleksandr-mykhailenko:
This one was closed. Need updates. Thanks

SunMar:
`webmozart/assert` is now upgradable, PR #960 was merged and tagged as [v4.5.2](https://github.com/mailgun/mailgun-php/releases/tag/v4.5.2).

PR #950, opened 2026-03-18
Update webmozart/assert dependency and remove strict types

aaronbauman:
addresses https://github.com/mailgun/mailgun-php/issues/948

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "php-http/discovery": "^1.19",
         "php-http/multipart-stream-builder": "^1.1.2",
         "psr/http-client": "^1.0",
-        "webmozart/assert": "^1.9.1"
+        "webmozart/assert": "^1.9.1 || ^2.0"
     },
     "require-dev": {
         "ergebnis/composer-normalize": "^2.43",
--- a/src/Assert.php
+++ b/src/Assert.php
@@ -16,18 +16,26 @@ use Mailgun\Exception\InvalidArgumentException;
 /**
  * We need to override Webmozart\Assert because we want to throw our own Exception.
  *
+ * Webmozart\Assert version 2.0 is required for PHP 8.4 support, but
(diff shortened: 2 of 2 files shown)


CircuitCipher:
Hoping to see this merged as I have a client that really needs mailgun to work due to limitations on their host and external SMTP and this old library dependency is preventing me from using the drupal/mailgun module.

aaronbauman:
Yeah, i've banged on this for a few hours now across 2 different PRs and can't figure out how to achieve compatibility for both versions.

I'll be running with the patched version, but someone else should feel free to pick this up.

CircuitCipher:
Hopefully maybe you'll get some support from the mailgun org

CircuitCipher:
@aaronbauman It looks like the implementation of the reportInvalidArgument method is the same in the mailgun implementation as it is in the parent Webmozart Assert.  Is it possible to just delete the override so that it doesn't fail the build and just let the class use the parent version?

smoench:
Can you try following:
```php
namespace Mailgun;

use Mailgun\Exception\InvalidArgumentException;

if (version_compare(\Composer\InstalledVersions::getVersion('webmozart/assert'), '2.0.0', '>=')) {
    final class Assert extends \Webmozart\Assert\Assert
    {
        protected static function reportInvalidArgument(string $message): never
        {
            throw new InvalidArgumentException($message);
        }
    }
} else {
    final class Assert extends \Webmozart\Assert\Assert
    {
        protected static function reportInvalidArgument($message): void
        {
            throw new InvalidArgumentException($message);
        }
    }
}

aaronbauman:
Nope. The static analysis is simply too aggressive.

aaronbauman:
Seems like a major version release for mailgun-php would be appropriate, dropping support for PHP < 8

This would allow significant reduction in the tooling and testing coverage, as well as updates for packages like webmozart/assert.

PR #959, opened 2026-07-13
support webmozart/assert ^2.0 (#948)

### LINKED ISSUE #948: Please update webmozart/assert requirement

SunMar:
**Support `webmozart/assert:^2.0`**
Add support for `webmozart/assert:^2.0` by converting the `Mailgun\Assert` class to a proxy class that forwards to, rather than extends from, `Webmozart\Assert\Assert`. Type hints that are not compatible with PHP 7.4 prevents overriding `reportInvalidArgument()` in a way that's compatible with `^1.0` and `^2.0` at the same time. This resolves #948. Fully backwards compatible, the proxy class still throws `Mailgun\Exception\InvalidArgumentException`. Only the stack trace origin will be from `__callStatic()` instead of `reportInvalidArgument()`. The original exception is still preserved through `$previous`.

**Added PHP 8.5 to the test workflow**
PHP 8.5 has been added as a target for the test workflow. Test results are now also parsed using [dorny/test-reporter](https://github.com/dorny/test-reporter).

**Added Roave BC Check**
Roave Backwards Compatibility Check was a no-op job in the workflow, it now runs the tool to check for breaking changes. it flags the `Assert` change which was explained above. It also flags `HttpApi::hydrateResponse()` but that should be an internal method which is safe to change.

**Updated static analysis tooling**
* _Psalm (`vimeo/psalm`)_: Bumped from `^4.0 || ^5.25` to `^5.25 || ^6.0`. The older `^5.25` was kept for when running in a PHP 7.4 / 8.0 environment). `psalm.xml` was updated to work with the new version.

* _PHPStan (`phpstan/phpstan`)_: Bumped to `^2.2`. It was pinned in the Github workflow to the very old version `0.12.92`. It now runs the latest version and is available locally.

* _PHP-CS-Fixer (`friendsofphp/php-cs-fixer`)_: Bumped to `^3.95`. It was pinned in the Github workflow to the very old version `3.9.5`. It now runs the latest version and is available locally.

* _PHP-Compatibility (`phpcompatibility/php-compatibility`)_: Changed to `dev-develop`, there has been no official release since 2019, but it is still actively being maintained in the `develop` branch. Additionally the `post-install-cmd` and `post-update-cmd` directives have been superseded by the plugin `dealerdirect/phpcodesniffer-composer-installer`, which is pulled in as a transitive dependency and will automatically handle installing the standards.

**Removed unused dev dependencies**
* _PHP_CodeSniffer (`squizlabs/php_codesniffer`)_: I could not find any direct references to this. The pipeline is using _PHP-CS-Fixer_. It's still pulled in as a transitive dependency of `phpcompatibility/php-compatibility`. It's worth considering to replace _PHP-CS-Fixer_ with _PHP_CodeSniffer_ as that package also provides a code fixer command (`phpcbf`), but this PR just sticks to the tooling already in-place.
* _Symfony HttpClient (`symfony/http-client`)_: No usages of this package and it was only a dev dependency, remove for being obsolete.

**Fixed static analysis errors**
All the errors produced by Psalm and PHPStan have been resolved, including the errors previously suppressed in `psalm.baseline.xml`. Most changes are just some tightening of type hints and code logic to satisfy Psalm and PHPStan, but do not introduce any functional changes (f.ex. changing nullables to non-nullable in cases where a NULL value is impossible already in the current code). Three errors appeared to be genuine (though edge-case) bugs:

* `HttpApi::handleErrors()`: The `case` statement for for 500 errors was unreachable, this would evaluate to `true` or `false`, neither of which can match `$statusCode`. Fixed by moving it to a proper `if()` statement in the `default:` case, so now it will throw an `HttpServerException` like intended, instead of an `UnknownErrorException`.
* `Pagination::getPaginationUrl()`: The callers could pass a NULL value for `$url`, this would trigger an exception because the argument is typed as non-nullable, that however doesn't seem to be in line with intended behavior due to the `if (empty($url)) { return null; }` statement. The method now returns NULL if no URL is provided.
* `BatchMessage::finalize()`: If the `NoopHydrator` is used, then `$response` is a `ResponseInterface` and not a `SendResponse` object. This would trigger an exception because `ResponseInterface` doesn't have a `getId()` method. A guard was added to prevent the exception in that use case. The caveat is that without a hydrator `getMessageIds()` will return an empty array, even if messages are sent.


SunMar:
@oleksandr-mykhailenko I was on holiday and wasn't able to immediately update the PR. Since this one was closed I've opened a new PR #960 which has the CI check resolved.

